PR #4, opened 2026-08-02
Agent/flint3 usb fixes

DuskyProjects:
## Summary

This PR fixes USB bring-up on the GL.iNet Flint 3 (GL-BE9300/IPQ5332).

The existing IPQ5332 device tree still describes USB using the older Qualcomm wrapper layout with a nested `snps,dwc3` child. The Linux 6.18 Qualcomm DWC3 driver expects a flattened controller node, so the controller does not bind and initialize correctly with the current layout.

The Flint 3 also requires two board-specific steps that were missing:

* GPIO 16 must be configured high to provide power to the external USB port.
* The shared USB/PCIe PHY mux must be switched to the USB path through the IPQ5332 TCSR register before the USB controller is reset.

## Changes

This PR:

* Flattens the IPQ5332 DWC3 device-tree node for the current `qcom,snps-dwc3` driver.
* Preserves both the existing `usb` and `usb_dwc` labels so current board DTS references continue to work.
* Expands the controller register range to cover the complete 1 MiB USB block, including the Qualcomm QSCRATCH registers at offset `0xf8800`.
* Moves the DWC3 core interrupt, PHY references, interconnects, clocks, and existing Synopsys quirks onto the flattened controller node.
* Adds the missing `dwc_usb3` interrupt and uses the interrupt ordering required by the Qualcomm binding.
* Adds a `qcom,multiplexed-phy` device-tree property.
* Updates the Qualcomm DWC3 driver to select the USB PHY path through the IPQ5332 TCSR USB/PCIe mux before asserting the controller reset.
* Uses `regmap_update_bits()` so only the USB-selection bit is changed and all unrelated TCSR bits are preserved.
* Applies the existing `usb_pins` pinctrl state to the Flint 3 USB controller, driving GPIO 16 high to power the port.

## Hardware details

The USB/PCIe PHY mux is controlled through:

* TCSR compatible: `qcom,tcsr-ipq5332`
* Register offset: `0x10540`
* USB selection mask: bit 0

The mux write is only performed when the board declares `qcom,multiplexed-phy`, so other IPQ5332 boards are not affected.

## Validation

Both kernel patches were checked with:

```sh
patch --dry-run --fuzz=0
```

against the exact Linux 6.18.39 source files used by this tree.

The final branch contains only:

* The device-tree binding patch.
* The IPQ5332 DWC3 and PHY-mux patch.
* The three required Flint 3 board DTS properties.

A full OpenWrt build and physical USB-device enumeration test should still be completed before the PR is merged.

### DIFF
--- a/target/linux/qualcommbe/dts/ipq5332-gl-be9300-usb3.dts
+++ b/target/linux/qualcommbe/dts/ipq5332-gl-be9300-usb3.dts
@@ -1,0 +1,92 @@
+// SPDX-License-Identifier: (GPL-2.0+ OR BSD-3-Clause)
+
+/*
+ * Keep the base Flint 3 board description intact and override only the
+ * USB path.  This makes the USB3 enablement easy to review separately from
+ * the already-tested Ethernet, Wi-Fi, eMMC and thermal board definitions.
+ */
+#include "ipq5332-gl-be9300.dts"
+
+/ {
+	usb_fixed_5p0: regulator-usb-5v {
+		compatible = "regulator-fixed";
+		regulator-name = "usb-fixed-5v";
+		regulator-min-microvolt = <5000000>;
+		regulator-max-microvolt = <5000000>;
+		regulator-boot-on;
+		regulator-always-on;
+	};
+};
+
+/*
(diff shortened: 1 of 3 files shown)


DuskyProjects:
Closed due to missing the driver, completely my fault.

PR #5, opened 2026-08-02
qualcommbe: gl-be9300: retry incomplete country setup

DuskyProjects:
# Title

qualcommbe: gl-be9300: retry incomplete country setup

# Description

## What changed

* require all three Flint 3 `wifi-device` sections before country initialization is considered complete
* regenerate the wireless configuration when fewer than three radio sections exist
* treat blank, malformed, and `00` country values as unset
* read the factory regulatory country from the ART partition
* apply the ART country only to radios that do not already have a valid country configured
* commit the wireless configuration and verify every radio afterward
* return a nonzero status when initialization is incomplete so OpenWrt retains the UCI-defaults script and retries it on the next boot

## Why

The existing `95-gl-wireless-country` script exits successfully on every failure path.

OpenWrt removes a UCI-defaults script after it exits successfully. If the script runs before the expected wireless sections are available, its radio loop can process nothing, report that the country was already set, and then be permanently removed.

The affected router is then left with blank country fields. Because 6 GHz requires a valid regulatory country, hostapd may refuse to start and Wi-Fi remains unavailable until the user manually selects a country.

This has been reported on at least two GL-BE9300 units. On one affected US unit, the ART partition correctly contains `US`, but all three radio country fields remained blank until they were manually configured.

## Result for users

On a normal first boot, the script now:

1. Confirms that all three radio sections exist.
2. Reads the factory country from ART.
3. Fills only blank or invalid radio country fields.
4. Commits the configuration.
5. Verifies that every radio now has a valid country.
6. Exits successfully only after initialization is complete.

If wireless configuration is temporarily incomplete, the script is retained and retried on the next boot instead of being silently deleted and leaving Wi-Fi permanently broken.

Existing valid country settings are preserved and are not overwritten.

## Validation

* Bash syntax check passed
* Dash syntax check passed
* BusyBox ash syntax check passed
* `git diff --check` passed
* only `95-gl-wireless-country` is changed

A fresh-install hardware test is still required.

### DIFF
--- a/target/linux/qualcommbe/ipq53xx/base-files/etc/uci-defaults/95-gl-wireless-country
+++ b/target/linux/qualcommbe/ipq53xx/base-files/etc/uci-defaults/95-gl-wireless-country
@@ -8,19 +8,44 @@
 # Without a country set, hostapd refuses to start on 6 GHz ("Invalid
 # country_code '00'"), so the radios stay down on a fresh install.
 #
-# Every exit path that leaves the country unset is logged: a silent failure
-# here looks like "Wi-Fi is broken" with nothing in the log to explain it.
+# This is a uci-defaults script. OpenWrt deletes it only when it exits 0.
+# Return non-zero whenever initialization is incomplete so it is retained
+# and retried on the next boot instead of silently becoming a one-shot no-op.
 
 . /lib/functions.sh
 . /lib/functions/system.sh
 
 LOG_TAG=gl-wireless-country
+EXPECTED_RADIOS=3
 
-warn_manual() {
+log_warn() {
 	logger -t "$LOG_TAG" -p daemon.warn "$1"
-	logger -t "$LOG_TAG" -p daemon.warn \
-		"radios may stay down (6 GHz needs a country). Set one manually, e.g.:" \
-		"uci set wireless.radio0.country=US; uci commit wireless; wifi"
+}
+
+retry_later() {
+	log_warn "$1"
+	log_warn "regulatory initialization incomplete; keeping this script for retry"
+	exit 1
+}
+
+list_wifi_devices() {
+	uci -q show wireless |
+		sed -n "s/^wireless\.\([^.]*\)=wifi-device$/\1/p"
+}
+
+normalize_country() {
+	printf '%s' "$1" |
+		tr '[:lower:]' '[:upper:]' |
+		tr -dc 'A-Z0-9'
+}
+
+valid_country() {
+	[ "${#1}" = 2 ] || return 1
+	[ "$1" != "00" ] || return 1
+	case "$1" in
+		*[!A-Z]*) return 1 ;;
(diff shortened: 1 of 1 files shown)


MNeroba:
This appears to have been superseded by later changes already present in `flint3-be9300`.

`ad30958f9f` (`qualcommbe: ipq53xx: set regulatory country on late-registering radios`) addresses the incomplete/late radio initialization directly through an `ieee80211` hotplug handler, so a late PHY can receive its ART country during the same boot rather than retaining the UCI-defaults script for the next boot.

`5a1620b1ba` (`qualcommbe: ipq53xx: don't treat country='00' as a country`) also covers the `country='00'` case from this PR and was verified on hardware.

The current boot sequence already runs `wifi config` before `kmodloader` and `uci_apply_defaults`, while the hotplug handler covers PHYs appearing afterwards. The newer implementation also deliberately avoids retrying a permanently invalid/unprogrammed ART state on every boot.

Because the current branch now covers the underlying failure modes more directly, I don't think rebasing this PR would add useful functionality. @DuskyProjects, could you please mark/close this PR as superseded by `ad30958f9f` and `5a1620b1ba`?

perceival:
Thanks for tracking this down — the uci-defaults exit-0-deletes-the-script race is a real bug and worth fixing.

This PR's diff can't apply cleanly anymore: `flint3-be9300` already picked up an independent fix (23d2fd409) for a different bug in the same file (treating `country='00'`, which `wifi config` writes for the 6 GHz radio by default, as invalid rather than just "unset" — `'00'` makes hostapd refuse the whole phy, not just 6 GHz).

Rather than a raw merge, I've hand-folded your actual contribution — wait for all 3 wifi-device sections to exist, regenerate via `wifi config` once if short, and return exit 1 (retain + retry next boot) on any incomplete/failed step instead of silently exiting 0 — on top of the existing `country_is_set`/'00'-rejection/ART-offset logic, in a75f06ba2f (pushed directly to `flint3-be9300`).

Verified before merging:
- 11-scenario mocked test harness (fresh install, the exact stuck-radio race this PR describes, partial/stale countries, ART missing/invalid, uci set/commit failures) — all behave as expected.
- Traced `wifi config`'s `radio_exists()` in `mac80211.uc` to confirm it's additive and won't clobber already-configured radios on the re-run.
- Traced `/etc/init.d/boot`'s uci-defaults runner to confirm `exit 1` retains-and-retries without blocking boot, and that the outer `uci commit` means no partial-write hazard even if the script's own `uci commit wireless` fails.
- Confirmed near-zero risk to the two already-deployed units (ap2, ap3): both already have valid countries on all 3 radios, so this exits at the "already set" check before touching ART or UCI at all.

Closing this in favor of a75f06ba2f, which carries the same fix. Thank you for the report and the fix — flagging two real units affected made this easy to prioritize.

PR #11, opened 2026-08-31
rtl837x: add DSA port mirroring

MNeroba:
## Summary

Wire the existing RTL8372 mirror block into the DSA `port_mirror_add` and
`port_mirror_del` callbacks.

The implementation uses the existing RTL837x RTK/DAL mirror API, which
provides one monitor port and independent RX/TX source masks.

## Why this is correct

The DSA callback signatures and `mirror->ingress` lifecycle match the kernel
API. DSA port indices are physical RTL837x port numbers in this driver, so
the source masks use `BIT(port)`. The DAL validates the monitor/source
relationship and programs the existing RTL8373 mirror registers; no raw
register programming or undocumented feature is introduced.

The RTL837x mirror block has one global direction selector bit. The
implementation programs `rx_tx_sel` from the requested direction (`RX_DIR`
for ingress and `TX_DIR` for egress), and rejects a second rule in the
opposite direction with `-EOPNOTSUPP` and an extack instead of returning
success while mirroring nothing. A different monitor port is rejected with
`-EBUSY`, and using the CPU port as the monitor is rejected as unsupported.

Per-source, per-direction reference counts prevent deleting one duplicate
rule from disabling another rule that still requires the same hardware mask
bit. Shadow state is updated only after successful hardware programming. The
last-rule delete disables the block and clears both masks; if either the
disable or clear operation fails, the shadow configuration is retained and a
best-effort rollback restores the previous hardware configuration. Setup and
teardown reset the mirror shadow state. CPU-port monitoring, RSPAN, sampling,
ACL mirroring, and multiple monitor groups remain outside the scope.

## Validation

  resolution; the final diff is limited to the two RTL837x mirror files.
- `git diff --check`: passed.
  warnings for the final patch.
- Cross-checked DSA callback signatures and the RTL8373 mirror DAL/register
  definitions.
- The final-delete failure handling was tightened in commit `c5fb1b9ee9`.
- No package build or hardware validation is claimed.

This PR remains Draft pending maintainer review and hardware validation.

## Requested hardware validation

Please verify ingress-only and egress-only `tc matchall` rules, multiple
same-direction sources including duplicate registration/deletion, rejection
of mixed directions and a second monitor port, CPU-port rejection, final-rule
deletion, teardown and reload, and behavior after an injected MDIO/SMI
failure.

### DIFF
--- a/package/kernel/rtl837x/src/rtl837x_common.h
+++ b/package/kernel/rtl837x/src/rtl837x_common.h
@@ -132,6 +132,8 @@ struct rtk_gsw {
 	int mirror_port;
 	u32 mirror_rx_mask;
 	u32 mirror_tx_mask;
+	bool mirror_direction_valid;
+	bool mirror_ingress;
 	struct net_device *ethernet_master;
 	struct sfp_bus *sfp_bus;
 
--- a/package/kernel/rtl837x/src/rtl837x_common.h
+++ b/package/kernel/rtl837x/src/rtl837x_common.h
@@ -127,6 +127,11 @@ struct rtk_gsw {
 	/* Hardware-indexed mask of ports with BR_ISOLATED enabled. */
 	u32 isolated_port_mask;
 	struct mutex isolation_lock;
+	bool port_enabled[RTK_MAX_NUM_OF_PORT];
+	/* RTL837x has one global mirror destination and separate RX/TX masks. */
+	int mirror_port;
+	u32 mirror_rx_mask;
+	u32 mirror_tx_mask;
 	struct net_device *ethernet_master;
 	struct sfp_bus *sfp_bus;
 
--- a/package/kernel/rtl837x/src/rtl837x_common.h
+++ b/package/kernel/rtl837x/src/rtl837x_common.h
@@ -132,6 +132,8 @@ struct rtk_gsw {
 	int mirror_port;
 	u32 mirror_rx_mask;
 	u32 mirror_tx_mask;
+	unsigned int mirror_rx_refcnt[RTK_MAX_NUM_OF_PORT];
+	unsigned int mirror_tx_refcnt[RTK_MAX_NUM_OF_PORT];
 	bool mirror_direction_valid;
 	bool mirror_ingress;
 	struct net_device *ethernet_master;
--- a/package/kernel/rtl837x/src/rtl837x_dsa_ops.c
+++ b/package/kernel/rtl837x/src/rtl837x_dsa_ops.c
@@ -697,14 +697,11 @@ static int rtl837x_read_ethtool_stat(int port, rtk_stat_port_type_t counter,
 }
 
 static int rtl837x_mirror_set_config(int mirror_port, u32 rx_mask,
-					     u32 tx_mask)
+					     u32 tx_mask, bool ingress)
 {
 	rtk_port_mir_set_t mir = {
 		.mtp_port = mirror_port,
-		/* The RTL8372N reference configuration uses both masks with the
-		 * selector at its default value.
-		 */
-		.rx_tx_sel = TX_DIR,
+		.rx_tx_sel = ingress ? RX_DIR : TX_DIR,
 	};
 
 	mir.rx_pmsk.bits[0] = rx_mask;
@@ -722,6 +719,7 @@ static int rtl837x_port_mirror_add(struct dsa_switch *ds, int port,
 	u32 rx_mask = gsw->mirror_rx_mask;
 	u32 tx_mask = gsw->mirror_tx_mask;
 	bool active = rx_mask || tx_mask;
+	int rollback_ret;
 	int ret;
 
 	if (!rtl837x_valid_port(gsw, port) ||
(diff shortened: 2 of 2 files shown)


MNeroba:
Additional review note

Static/API confidence is approximately 90%: the patch uses the existing DSA mirror callbacks and the existing RTL8372N RTK/DAL mirror API, without introducing raw register programming. The one global monitor-port limitation is represented explicitly, source RX/TX masks are retained independently, and error paths preserve the software state.

The remaining uncertainty is silicon/runtime behavior: in particular, the TX_DIR selector used with both masks, packet direction semantics, CPU-port handling, and reconfiguration after a failed hardware operation. Please validate these points on a Flint 3 with ingress and egress mirror rules, multiple sources sharing one monitor port, rejection of a second monitor port, and removal of the final rule. Until that test, the implementation should remain a draft if maintainers prefer hardware confirmation before merge.

perceival:
The plumbing here is sound and I want this feature — the `port_mirror_add` /
`port_mirror_del` signatures match `include/net/dsa.h` on both 6.12 and 6.18,
`mirror->ingress` is set by the DSA core before the add callback so the del path
may rely on it, the `BIT(port)` masks are right because this driver uses DSA
indices as physical port numbers throughout, and the single-monitor `-EBUSY`
guard plus the `mirror_port = -1` reset in setup/teardown are correct. It also
can't wedge the CPU port: the core only accepts a user port as monitor, and
`dal_rtl8373_mirror_entry_set` additionally rejects a monitor present in either
mask.

One blocking issue. `rtl837x_mirror_set_config()` initialises
`.rx_tx_sel = TX_DIR` with the comment "uses both masks with the selector at its
default value". That comment doesn't hold: `rx_tx_sel` is written to a *single*
hardware bit — `dal_rtl8373_mirror.c` writes
`RTL8373_MIR_SET_CTRL_MIR_RX_TX_SEL_OFFSET` (bit 5 of 0x6048) — i.e. a direction
selector, not an "enable both". Pinned to `TX_DIR`, a
`tc ... matchall skip_sw action mirred ingress` rule programs `rx_pmsk`, returns
0, and mirrors nothing. Silent success is worse than an error here.

Either drive the bit from the requested direction, or reject the unsupported
direction with an extack message so the user learns why.

Two nits, non-blocking: a mid-sequence failure inside
`dal_rtl8373_mirror_entry_set` leaves the hardware masks out of sync with
`gsw->mirror_*` (no rollback), and the del path clears only the global enable
bit, leaving stale masks in the registers.

MNeroba:
Thanks, agreed — the selector is global and `TX_DIR` was incorrect for ingress mirroring. I pushed follow-up commit `007b4de9f2`.

The mirror configuration now sets `rx_tx_sel` from the requested direction (`RX_DIR` for ingress, `TX_DIR` for egress) and rejects a second rule in the opposite direction with `-EOPNOTSUPP` plus an extack message. This avoids a silent-success configuration that mirrors nothing when both directions are requested.

I also added best-effort restoration of the previous shadow configuration after a partial hardware write, clears both hardware masks when the final rule is removed, and resets the direction state during teardown. The global monitor-port limitation and the existing single-direction limitation are now explicit in the implementation.

Please re-check ingress-only and egress-only rules, multiple source ports in one direction, rejection of mixed directions, final-rule deletion, and a forced mid-sequence SMI failure.

MNeroba:
Rebased the PR onto the current `flint3-be9300` base and pushed commit `d8a3727694`; the reviewed direction fix is unchanged semantically.

The hardware detail from your review is now reflected explicitly: `rx_tx_sel` is programmed from the requested direction, and a second rule with the opposite direction is rejected with an extack because RTL837x has one global selector bit. The add/delete paths also restore the previous mask configuration after a failed read-modify-write where possible, clear masks when the last rule is removed, and reset the direction state during teardown.

Static confidence is about 97%: the DSA callback signatures, single-monitor guard, direction selector, and rollback state machine are all directly covered by the current kernel API and RTL8373 mirror DAL. The remaining check is hardware behavior on RTL8372N: ingress and egress capture separately, multiple same-direction sources, and teardown/reload after an injected MDIO failure.

perceival:
Reviewed against 6.18.39. The mechanics check out: callback signatures match `net/dsa/user.c`, `mirror->ingress` is set before `port_mirror_add` and preserved for `port_mirror_del` so reading it in del is safe, both callbacks run under rtnl so the unlocked shadow state is fine, and the port numbering is right (DSA index == physical port here, and the RTL8373 L2P table is identity).

Two things I'd like addressed before merge.

**1. No per-rule refcount on the source-port mask — deleting one of two rules kills both.**

`rx_mask |= BIT(port)` is idempotent, so a second matchall on the same port and direction succeeds without recording that it exists.

```
tc filter add dev lan1 ingress prio 1 matchall action mirred egress mirror dev lan3
tc filter add dev lan1 ingress prio 2 matchall action mirred egress mirror dev lan3
tc filter del dev lan1 ingress prio 1
```

The del clears `BIT(lan1)` and reprograms the hardware, so mirroring for the still-installed prio-2 filter is silently dead while `tc` still reports it as offloaded.

**2. `mirror->to_local_port` can be the CPU port, and it is currently accepted.**

`dsa_user_add_cls_matchall_mirred()` sets `to_dp = dp->cpu_dp` whenever the mirred target is a foreign netdev (confirmed at `net/dsa/user.c`, the "Handle mirroring to foreign target ports as a mirror towards the CPU" path). `rtl837x_valid_port()` includes `cpu_port`, so we happily program the CPU port as the RTL8373 monitor port.

Because this switch uses `DSA_TAG_PROTO_VSC73XX_8021Q`, a mirrored copy arriving on the CPU port carries no valid tag_8021q RX VID, so `vsc73xx_rcv()` logs "Couldn't decode source port" and drops every copy — while the mirrored traffic still burns CPU-port bandwidth and EDMA ring slots.

```
tc qdisc add dev lan1 clsact
tc filter add dev lan1 ingress matchall action mirred egress mirror dev dummy0
```

Suggest rejecting with `-EOPNOTSUPP` when `to_local_port == gsw->cpu_port`.

**3. (minor)** If `rtk_mirror_set_en(DISABLED)` fails on the last-rule delete, the early return leaves the shadow masks populated. The tc filter is already gone at that point, so nothing will ever clear them. Recoverable by re-adding the same rule, so not a blocker.

MNeroba:
Addressed the follow-up review in `91f666bd3d`:

- mirroring now tracks each source independently per direction, so deleting one of several rules does not disable the remaining source;
- CPU-port monitor destinations return `-EOPNOTSUPP` with an extack instead of being partially accepted;
- the final delete attempts to disable the hardware and releases the corresponding software state.

The body and validation notes are updated; direction and rollback behavior still need hardware testing before this Draft is ready.

MNeroba:
(`a75f06ba2f`) and pushed `c5fb1b9ee9`.

In addition to the already reviewed per-source reference counts and CPU-port
monitor rejection, the final-delete path now keeps the shadow state when
disabling or clearing the hardware fails. A failed clear attempts to restore
the previous masks and re-enable the mirror, so a failed callback does not
silently lose the software ownership state.

`git diff --check` and checkpatch pass. This remains Draft because the
RTL8373 mirror direction/packet behavior and injected-SMI recovery still need
hardware validation.

PR #12, opened 2026-08-31
rtl837x: check switch setup errors

MNeroba:
## Summary

`rtl837x_setup()` currently ignores the return values from
`rtk_l2_table_clear()`, `rtk_l2_aging_set(300)`, and
`rtk_stat_global_reset()`.

All three APIs report MDIO/hardware errors. If one fails, setup continues into
VLAN and DSA registration, leaving a switch that was reported as initialized
although one of its basic subsystems may not be configured.

The surrounding setup code already checks each RTK API result and converts
failures with `rtl837x_to_errno()`. This patch applies the same existing error
handling to the three unchecked operations, without changing their arguments
or success-path ordering.

The RTK headers document non-`RT_ERR_OK` results for all three calls, including
SMI access failures. The DAL implementations propagate the underlying register
operation result, so ignoring it loses the only setup-time indication that the
switch could not be initialized reliably.

## Scope and risk

* Success behavior is unchanged.
* On an MDIO or hardware error, setup now aborts before DSA registration.
* No VLAN, tagger, forwarding, or packet-path behavior is changed.
* No Flint 3 hardware is required to establish the missing error checks;
  hardware testing can still confirm the failure reporting path.

## Validation

* `scripts/checkpatch.pl --no-tree -`: 0 errors, 0 warnings.
* `git diff --check`.
* Verified the three functions return `rtk_api_ret_t` in their public headers
  and are mapped to the RTL8373 DAL implementation.

The full OpenWrt package build could not run on this macOS host because the
checkout requires GNU make newer than Apple's Make 3.81; the command stopped
before compilation.

## Review request

Please verify that these setup calls are expected to be fatal, especially the
LUT clear operation, and that no board relies on continuing after a reported
failure. If one operation is intentionally best-effort, it should be kept
explicitly separate rather than silently ignoring its return value.

## Confidence

High (approximately 0.95) for the correctness of checking and propagating the
existing return codes. Runtime confidence is high for the success path because
the calls and ordering are unchanged; the failure path still deserves a
maintainer review on real hardware or an injected MDIO-error test.

### DIFF
--- a/package/kernel/rtl837x/src/rtl837x_dsa_ops.c
+++ b/package/kernel/rtl837x/src/rtl837x_dsa_ops.c
@@ -277,9 +277,17 @@ static int rtl837x_setup(struct dsa_switch *ds)
 	if (ret)
 		return rtl837x_to_errno(ret);
 
-	rtk_l2_table_clear();
-	rtk_l2_aging_set(300);
-	rtk_stat_global_reset();
+	ret = rtk_l2_table_clear();
+	if (ret)
+		return rtl837x_to_errno(ret);
+
+	ret = rtk_l2_aging_set(300);
+	if (ret)
+		return rtl837x_to_errno(ret);
+
+	ret = rtk_stat_global_reset();
+	if (ret)
+		return rtl837x_to_errno(ret);
 
 	ret = rtk_vlan_reset();
 	if (ret)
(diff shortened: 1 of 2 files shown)


MNeroba:
Follow-up commit `a5493460ec` adds propagation of `extra-init` register-write failures.

Why this is safe: `of_extra_init()` already returns `int`, but the result of `rtl8373_setAsicRegBits()` was discarded. The new path logs the exact register, mask, value, and SDK error, then stops `rtl8372n_hw_init()` before VLAN setup when a write fails. The successful path and all register values are unchanged.

This is failure-path hardening, not a new hardware feature. The commit passes `checkpatch.pl` with 0 errors and 0 warnings. No Flint 3 runtime hardware was available, so please verify whether a failed `extra-init` write should be fatal for probe or remain best-effort.

PR #18, opened 2026-08-31
rtl837x: handle GPIO controller registration errors

MNeroba:
## Summary

When the device tree requests `gpio-controller`, `rtl837x_dsa_probe()` calls `rtl837x_gpiochip_init()` but previously ignored its return value. A failed `devm_gpiochip_add_data()` therefore left the DSA switch registered while the requested GPIO provider was unavailable.

Check the return value, unregister the already registered DSA switch, release the Ethernet master reference, and return the original errno. This preserves `-EPROBE_DEFER` and other failures for the driver core. Device trees without `gpio-controller` and successful GPIO registration follow the existing path unchanged.

## Validation

- `git diff --check`
- checkpatch on the proposed diff: no errors or warnings
- No Flint 3 hardware was available for runtime testing

Confidence: 97% based on the local DSA registration/teardown lifecycle and the GPIO provider API contract.

Please verify on hardware or with representative device trees covering:

- no `gpio-controller` property;
- successful GPIO controller registration;
- GPIO registration returning `-EPROBE_DEFER`, `-ENOMEM`, or another error after DSA registration.

### DIFF
--- a/package/kernel/rtl837x/src/rtl837x_mdio.c
+++ b/package/kernel/rtl837x/src/rtl837x_mdio.c
@@ -875,8 +875,18 @@ static int rtl837x_dsa_probe(struct mdio_device *mdiodev)
 	}
 
 #ifdef CONFIG_GPIOLIB
-	if (of_property_read_bool(np, "gpio-controller")) 
-		rtl837x_gpiochip_init(gsw);
+	if (of_property_read_bool(np, "gpio-controller")) {
+		ret = rtl837x_gpiochip_init(gsw);
+		if (ret) {
+			dev_err_probe(dev, ret,
+				      "failed to register GPIO controller\n");
+			rtl837x_dsa_unregister(gsw);
+			if (master)
+				dev_put(master);
+			devm_kfree(dev, gsw);
+			return ret;
+		}
+	}
 #endif /* CONFIG_GPIOLIB */
 
 	rtl837x_sfp_probe(gsw);


MNeroba:
Request for review: this patch covers the failure path that was previously invisible after `dsa_register_switch()` succeeded. If the DT declares a GPIO provider, probe success should imply that provider was registered; otherwise consumers can receive a false-positive switch probe and fail later. The cleanup uses the existing `rtl837x_dsa_unregister()` path before releasing `master`, and returns the exact GPIO registration errno so deferred dependencies remain retryable. Confidence is 97% from source-level lifecycle analysis; please confirm the teardown ordering and deferred-probe case on the target kernel.

MNeroba:
Superseded by consolidated PR #44. GPIO-controller registration failure handling is retained with the equivalent SFP registration unwind in one auxiliary-probe review.

perceival:
The `dev_err_probe`-style reporting is welcome, but this makes a non-fatal
condition fatal for the whole switch.

`rtl837x_gpiochip_init()` ends in `devm_gpiochip_add_data()`, which can fail for
reasons unrelated to the data path — a `base` collision from the DT `base`
property, or gpiolib resource exhaustion. Today that failure is ignored and the
switch keeps working without GPIO; with this patch it unregisters DSA and fails
probe, taking the router's LAN down with it.

Suggested shape: log with `dev_err_probe()` and continue, or propagate only
`-EPROBE_DEFER` and swallow the rest.

Secondary, and shared with #19: the new error path calls `devm_kfree(dev, gsw)`
while `dev_set_drvdata(dev, gsw)` and the module-global `rtl_gbl_priv = gsw`
still point at that object, and both `devm_regmap_init()` contexts still
reference it until devres unwinds. That mirrors the pre-existing
`rtl8372n_hw_init` / `rtl837x_dsa_register` error paths, so it isn't a new class
of bug, but it would be better to drop the redundant `devm_kfree` and clear the
stale pointers than to propagate the pattern. #28 (which I'm merging) fixes the
global for the existing paths.

PR #19, opened 2026-08-31
rtl837x: handle SFP probe errors

MNeroba:
## Summary

Handle failures from the optional SFP bus setup instead of silently continuing with a partially initialized switch.

When an SFP bus is configured but rtl837x_sfp_probe() cannot attach it, the probe now:

- reports the original errno with dev_err_probe();
- unregisters the DSA switch;
- releases the Ethernet master reference;
- returns the error to the driver core.

The no-SFP path remains successful.

The helper also clears gsw->sfp_bus when sfp_bus_add_upstream() fails. The SFP core manages its own bus reference on that failure, while the caller drops the reference returned by sfp_bus_find_fwnode(). Retaining the pointer after a failed add could leave remove() with a stale pointer.

## Why this is correct

The Linux SFP core treats a missing SFP reference as an optional no-bus case, while non-zero results from bus discovery or upstream registration are actual errors. sfp_bus_add_upstream() drops its temporary reference when registration fails, so the driver must not retain the failed bus pointer.

This change is limited to probe error propagation and lifetime cleanup. It does not change SFP capability selection, link modes, or normal attach/detach behavior.

## Validation

- git diff --check

## Confidence

97%. The control flow follows the upstream SFP bus reference/error contract. Please verify on hardware or with fault injection that:

1. no SFP property still allows successful probe;
2. a valid SFP bus still attaches and removes normally;
3. an invalid or deferred SFP dependency fails probe with the original errno;
4. an upstream-registration failure does not leave a bus pointer for remove() to use.

### DIFF
--- a/package/kernel/rtl837x/src/rtl837x_mdio.c
+++ b/package/kernel/rtl837x/src/rtl837x_mdio.c
@@ -665,6 +665,8 @@ static int rtl837x_sfp_probe(struct rtk_gsw *gsw)
 	gsw->sfp_bus = bus;
 
 	ret = sfp_bus_add_upstream(bus, gsw, &sfp_ops);
+	if (ret)
+		gsw->sfp_bus = NULL;
 	sfp_bus_put(bus);
 
 	return ret;
@@ -879,7 +881,15 @@ static int rtl837x_dsa_probe(struct mdio_device *mdiodev)
 		rtl837x_gpiochip_init(gsw);
 #endif /* CONFIG_GPIOLIB */
 
-	rtl837x_sfp_probe(gsw);
+	ret = rtl837x_sfp_probe(gsw);
+	if (ret) {
+		dev_err_probe(dev, ret, "failed to attach SFP bus\n");
+		rtl837x_dsa_unregister(gsw);
+		if (master)
+			dev_put(master);
+		devm_kfree(dev, gsw);
+		return ret;
+	}
 
 	rtl837x_debug_proc_init(gsw);
 	rtl837x_status_check_work_init(gsw);


MNeroba:
Request for focused review and runtime validation.

The important lifetime point is the SFP core contract: sfp_bus_add_upstream() takes a bus reference and releases its registration reference if setup fails, while the caller still releases the reference returned by sfp_bus_find_fwnode(). Keeping gsw->sfp_bus non-NULL after a non-zero return could therefore leave rtl837x_dsa_remove() with a stale pointer.

The probe now propagates a real SFP discovery/registration error and unwinds the already registered DSA switch. A missing SFP remains a successful optional configuration.

Confidence: 97% based on the upstream SFP bus implementation and a minimal one-file diff. Please check no-SFP, valid SFP, deferred/invalid dependency, and forced registration-failure paths.

MNeroba:
Superseded by consolidated PR #44. SFP registration failure handling is retained with the equivalent GPIO-controller unwind in one auxiliary-probe review.

perceival:
Hunk 1 is a real fix and I'd take it on its own: if `sfp_bus_add_upstream()`
fails it has already dropped its own reference, so leaving `gsw->sfp_bus` set
means `rtl837x_dsa_remove()` later calls `sfp_bus_del_upstream()` on a bus that
was never added. Clearing it is correct.

Hunk 2 is the problem — same objection as #18. `rtl837x_sfp_probe()` returns
non-zero only on a real SFP subsystem failure, and `sfp_bus_find_fwnode()`
returns NULL (not an ERR_PTR) when there is no `sfp` property, so BE9300 and
BE6500 always get `ret == 0` and are unaffected today. But on any RTL837x board
that does wire an SFP cage, an optional-cage failure would now tear down the
already-registered DSA switch and kill the entire LAN path. The unwind also
frees `gsw` while `drvdata` and `rtl_gbl_priv` still reference it.

Could you split this: keep hunk 1, and either drop hunk 2 or downgrade it to
`dev_warn` + continue? I'll merge the split immediately.

PR #21, opened 2026-08-31
rtl837x: free debugfs write buffers

MNeroba:
## Summary

Free the user-copy buffers allocated by the three debugfs register write handlers.

_sdsreg_rw_write(), _phyreg_mmd_rw_write(), and _reg_rw_write() each call memdup_user_nul() but previously leaked the returned buffer on every successful write and on several validation-error paths. This patch frees the buffer before every return after a successful allocation.

## Why this is correct

Linux documents memdup_user_nul() as returning a newly allocated user-data copy. The allocation is physically contiguous and must be released with kfree(). The handlers do not retain the buffer after parsing and register access, so freeing it at the end of each path is the correct lifetime.

The patch does not alter register values, accepted command syntax, return values, or the existing debugfs interface. It only closes the allocation lifetime on normal and early-error exits.

## Validation

- git diff --check
- Confirmed rtl837x_dbg.o is part of the rtl837x_dsa module

## Confidence

99%. This is a direct allocation/free lifetime fix with no hardware-dependent behavior change. Please review the three write handlers and, if possible, confirm repeated writes do not grow kernel memory usage under kmemleak or slab accounting.

### DIFF
--- a/package/kernel/rtl837x/src/rtl837x_dbg.c
+++ b/package/kernel/rtl837x/src/rtl837x_dbg.c
@@ -49,23 +49,29 @@ ssize_t _sdsreg_rw_write(struct file *filep, const char __user *ubuf,
 		return PTR_ERR(buf);
 	
 	if(buf[0] == 'w') {
-		if(sscanf(buf, "w %d %x %x %x", &sds_id, &page, &reg, &val) == -1)
+		if(sscanf(buf, "w %d %x %x %x", &sds_id, &page, &reg, &val) == -1) {
+			kfree(buf);
 			return -EFAULT;
-		else{
+		} else {
 			if (sds_id > 1)
+			{
+				kfree(buf);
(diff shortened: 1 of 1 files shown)


MNeroba:
Focused review request.

All three debugfs write handlers allocate with memdup_user_nul() and previously returned without kfree(), including their malformed-input paths. Because these handlers are part of the built rtl837x_dsa module, repeated writes can leak kernel heap memory indefinitely.

The patch adds only the missing frees and leaves command behavior unchanged. Confidence: 99%. Please inspect the normal and early-error paths, and if available use kmemleak or slab accounting while repeating writes to reg, phy_mmd, and sdsreg.

MNeroba:
Superseded by consolidated PR #45. Debugfs buffer lifetime handling is retained together with parser validation because both affect the same write-handler exit paths.

PR #22, opened 2026-08-31
rtl837x: validate debugfs command input

MNeroba:
## Summary

Validate the number and types of fields parsed by the debugfs register commands.

The three debugfs write handlers previously treated only sscanf() == -1 as a parse failure. Ordinary malformed or truncated input returns the number of successfully assigned fields (0..N), so commands such as "w" or "w 1" could use uninitialized register, page, or value variables. The handlers now require the exact number of fields for each command. The unsigned sds_id and port fields also use %u, matching their uint32_t destinations.

## Why this is correct

sscanf() reports the number of successful assignments; an input error is not the only invalid result. Requiring 4, 3, 2, or 1 assignments for the four command forms prevents register access with uninitialized arguments. This follows established upstream kernel patterns that validate sscanf() against the expected conversion count.

The change preserves the existing accepted command forms and return value for valid input. It only rejects incomplete or malformed commands that were previously processed unsafely.

This patch is intentionally independent of PR #21, which addresses the separate debugfs allocation cleanup.

## Validation

- git diff --check

## Confidence

99%. The fix is a direct parser-contract correction with no hardware behavior change. Please verify:

1. Valid sdsreg, phyreg_mmd, and reg read/write commands.
2. Empty and partial commands.
3. Out-of-range sds_id and port values.
4. Extra trailing fields, if the intended interface should reject them.

### DIFF
--- a/package/kernel/rtl837x/src/rtl837x_dbg.c
+++ b/package/kernel/rtl837x/src/rtl837x_dbg.c
@@ -49,7 +49,7 @@ ssize_t _sdsreg_rw_write(struct file *filep, const char __user *ubuf,
 		return PTR_ERR(buf);
 	
 	if(buf[0] == 'w') {
-		if(sscanf(buf, "w %d %x %x %x", &sds_id, &page, &reg, &val) == -1)
+		if (sscanf(buf, "w %u %x %x %x", &sds_id, &page, &reg, &val) != 4)
 			return -EFAULT;
 		else{
 			if (sds_id > 1)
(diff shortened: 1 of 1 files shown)


MNeroba:
Focused review request.

The previous checks used sscanf(...) == -1. For malformed input with zero or partial conversions, sscanf() returns 0..N, so uninitialized fields could reach the register access functions. The patch requires the exact conversion count and aligns %u with the uint32_t destinations.

This is modeled on upstream kernel code that validates sscanf() return counts explicitly. Confidence: 99%. Please test valid commands and empty or partial commands with all three debugfs files. The extra trailing-field behavior is called out separately in the PR description for confirmation of the intended interface.

MNeroba:
Superseded by consolidated PR #45. Debugfs parser validation is retained together with buffer lifetime handling because both affect the same write-handler exit paths.

PR #23, opened 2026-08-31
rtl837x: handle CPU port status read errors

MNeroba:
## Summary

Handle failures while reading the CPU port status in the periodic SerDes recovery worker.

The worker previously ignored the return value of rtk_port_macStatus_get() and immediately consumed port_status.link. The Realtek DAL can return before populating the output structure when an SMI register read fails, so the worker could make a decision based on an uninitialized value and reset the SerDes spuriously.

On a status read error, the worker now emits a rate-limited warning, skips recovery for that iteration, and schedules the next periodic check. The status structure is also initialized defensively.

## Why this is correct

rtk_port_macStatus_get() explicitly reports SMI and driver errors. In the bundled RTL8373 DAL implementation, every register read is checked and the function returns immediately on failure, before all fields of rtk_port_status_t have been written. Ignoring that result is therefore unsafe.

A failed status read is not evidence that the CPU link is down. Skipping the reset avoids changing hardware state on an unknown status, while requeueing preserves the existing monitoring behavior. Successful reads follow the existing link detection and SerDes recovery path unchanged.

The warning is rate-limited because the worker polls once per second and a persistent SMI failure should not flood the kernel log.

## Validation

- git diff --check

## Confidence

98%. The bug is directly established by the ignored API error and the DAL early-return behavior. Please verify:

1. Normal CPU link-up polling remains unchanged.
2. A real CPU link-down event still follows the existing SerDes recovery path.
3. Injected or observed SMI/MDIO read failures produce a warning without a SerDes reset.
4. Polling resumes after a transient status read failure.

This PR is intentionally submitted for maintainer review and hardware/runtime validation.

### DIFF
--- a/package/kernel/rtl837x/src/rtl837x_mdio.c
+++ b/package/kernel/rtl837x/src/rtl837x_mdio.c
@@ -537,12 +537,19 @@ static void rtl837x_status_check_work_func(struct work_struct *work)
 {
 	struct rtk_gsw *gsw = container_of(work, struct rtk_gsw, status_check_work.work);
 
-	rtk_port_status_t port_status;
+	rtk_port_status_t port_status = { 0 };
+	int ret;
 
 	if (!gsw->ethernet_master)
 		return;
 
-	rtk_port_macStatus_get(gsw->cpu_port, &port_status);
+	ret = rtk_port_macStatus_get(gsw->cpu_port, &port_status);
+	if (ret) {
+		dev_warn_ratelimited(gsw->dev,
+				     "failed to read CPU port status: %d\n", ret);
+		goto reschedule;
+	}
+
 	if (!port_status.link)
 	{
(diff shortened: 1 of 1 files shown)


MNeroba:
Focused review request.

The worker used port_status.link after ignoring rtk_port_macStatus_get(). The bundled DAL returns immediately on an SMI register error, so the output structure is not guaranteed to be populated. The patch treats a failed read as unknown: it avoids a potentially false SerDes reset, logs the error with rate limiting, and keeps the periodic check alive.

Confidence: 98%. Please verify normal link-up/link-down behavior and, if possible, inject an SMI/MDIO read failure to confirm that the worker retries without changing SerDes mode.

MNeroba:
Superseded by consolidated PR #41. The status-read fix is retained there together with the related worker lifecycle changes, so reviewers can verify the complete path in one place.

PR #24, opened 2026-08-31
rtl837x: report SFP SerDes removal errors

MNeroba:
## Summary

Report failures while disabling the SFP SerDes during module removal.

rtl837x_sfp_module_remove() previously ignored the return value of rtk_sdsMode_set(). If the switch rejects the requested SerDes mode or the SMI transaction fails, teardown continues without any diagnostic.

The callback now records the return value and emits a device error when the hardware operation fails. The callback remains non-failing because the upstream SFP API defines module_remove as a void callback.

## Why this is correct

The Realtek API returns a status for rtk_sdsMode_set(), including hardware access failures. Ignoring that status hides a failed state transition and makes runtime diagnosis unnecessarily difficult.

The Linux SFP contract gives module_remove no return channel, so logging is the only non-invasive error reporting available at this boundary. Successful behavior is unchanged, and teardown is not blocked by a module-removal error.

This is the removal-side counterpart to the insertion error reporting proposed in PR #20.

## Validation

- git diff --check

## Confidence

97%. The return value is a documented part of the Realtek API and the SFP callback cannot propagate an error. Please verify:

1. Removing a working SFP module still disables the configured SerDes mode.
2. An injected or observed SMI/SerDes failure produces the new diagnostic.
3. SFP teardown completes normally even when the hardware transition fails.

### DIFF
--- a/package/kernel/rtl837x/src/rtl837x_mdio.c
+++ b/package/kernel/rtl837x/src/rtl837x_mdio.c
@@ -632,10 +632,13 @@ static int rtl837x_sfp_module_insert(void *upstream, const struct sfp_eeprom_id 
 static void rtl837x_sfp_module_remove(void *upstream)
 {
 	struct rtk_gsw *gsw = upstream;
+	rtk_api_ret_t ret;
 	dev_info(gsw->dev, "SFP module remove\n");
 
 	USE_SERDESMODE(1, SERDES_OFF);
-	rtk_sdsMode_set(1, gsw->sds1mode);
+	ret = rtk_sdsMode_set(1, gsw->sds1mode);
+	if (ret)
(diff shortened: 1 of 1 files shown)


MNeroba:
Focused review request.

module_remove() is a void SFP callback, so a failed rtk_sdsMode_set() cannot be returned to the SFP core. The patch preserves teardown while surfacing the failure with dev_err, which prevents silent SerDes state-transition failures.

Confidence: 97%. Please verify normal module removal and, if possible, an SMI/SerDes failure during removal. PR #20 covers the separate insertion callback, which can return an error.

MNeroba:
Superseded by consolidated PR #39. The SFP removal error reporting is retained in the canonical SFP SerDes transition review together with insertion handling and shadow-state rollback.

PR #25, opened 2026-08-31
rtl837x: report FDB flush errors

MNeroba:
## Summary

Report failures from the FDB flush operation used by the DSA fast-age callback.

rtl837x_port_fast_age() previously called rtk_l2_ucastAddr_flush() and discarded its return value. A failed SMI transaction or switch-side operation could therefore leave stale dynamic FDB entries while the kernel received no diagnostic.

The callback now preserves the return value and reports failures with dev_err.

## Why this is correct

The Realtek API declares rtk_l2_ucastAddr_flush() as returning rtk_api_ret_t, with RT_ERR_OK as the success result and hardware/API errors otherwise. The operation is the actual implementation of DSA fast-age for a user port, so a failure is operationally relevant.

The DSA port_fast_age callback is void and cannot report an errno to the bridge/DSA layer. Logging the error is therefore the available non-invasive handling at this boundary. The hardware operation is still attempted exactly as before, and successful behavior is unchanged.

## Validation

- git diff --check
- Verified the only changed file is package/kernel/rtl837x/src/rtl837x_dsa_ops.c

## Confidence

96%. The API return contract and the DSA callback signature are explicit. Please verify:

1. FDB entries are flushed normally after a port fast-age event.
2. An injected or observed switch/SMI failure emits the new diagnostic.
3. The callback remains safe and non-blocking when the flush fails.

### DIFF
--- a/package/kernel/rtl837x/src/rtl837x_dsa_ops.c
+++ b/package/kernel/rtl837x/src/rtl837x_dsa_ops.c
@@ -677,6 +677,7 @@ static void rtl837x_port_fast_age(struct dsa_switch *ds, int port)
 {
 	struct rtk_gsw *gsw = ds->priv;
 	rtk_l2_flushCfg_t cfg = { 0 };
+	rtk_api_ret_t ret;
 
 	if (!rtl837x_user_port(gsw, port))
 		return;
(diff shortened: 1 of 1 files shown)


MNeroba:
Focused review request.

DSA fast-age is implemented by rtk_l2_ucastAddr_flush(), whose Realtek API returns a status but was previously ignored. Because port_fast_age is a void DSA callback, this patch reports a failed flush without changing the normal operation or teardown behavior.

Confidence: 96%. Please verify dynamic FDB removal after fast-age and, if possible, force an SMI failure to confirm that the error is visible in the kernel log.

PR #26, opened 2026-08-31
rtl837x: report STP state errors

MNeroba:
## Summary

Report failures while programming STP state for DSA ports.

The port_disable and port_stp_state_set callbacks previously discarded the result of rtl837x_set_stp_state(). A failed Realtek register transaction could leave the hardware in its previous forwarding state while the kernel received no diagnostic.

The callbacks now log the returned error together with the affected port and requested state. The existing hardware operation and DSA callback behavior are otherwise unchanged.

## Why this is correct

rtl837x_set_stp_state() converts the Realtek MSTP API result into a Linux error, so its return value is meaningful and must not be silently discarded. The DSA port_disable and port_stp_state_set callbacks are both void, which prevents returning the error to the DSA core; logging is the non-invasive way to expose a hardware/software state mismatch at this boundary.

Successful transitions are unchanged. On failure, the code still attempts the requested operation and reports the failure rather than hiding it.

## Validation

- git diff --check
- Verified the only changed file is package/kernel/rtl837x/src/rtl837x_dsa_ops.c

## Confidence

96%. The Realtek API return path, Linux error conversion, and DSA callback signatures are explicit. Please verify:

1. Normal port enable/disable and bridge STP transitions.
2. An injected or observed SMI failure produces the new diagnostic.
3. The hardware state is checked after a failed STP write.

### DIFF
--- a/package/kernel/rtl837x/src/rtl837x_dsa_ops.c
+++ b/package/kernel/rtl837x/src/rtl837x_dsa_ops.c
@@ -657,20 +657,27 @@ static int rtl837x_port_enable(struct dsa_switch *ds, int port,
 static void rtl837x_port_disable(struct dsa_switch *ds, int port)
 {
 	struct rtk_gsw *gsw = ds->priv;
+	int ret;
 
 	if (!rtl837x_valid_port(gsw, port))
 		return;
 
 	gsw->port_enabled[port] = false;
-	rtl837x_set_stp_state(gsw, port, BR_STATE_DISABLED);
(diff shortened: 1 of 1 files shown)


MNeroba:
Focused review request.

Both DSA STP callbacks are void, but the underlying Realtek write returns an error through rtl837x_set_stp_state(). The patch keeps the operation unchanged and surfaces a failed register transaction with the port and requested state.

Confidence: 96%. Please verify normal STP transitions and, if possible, force an SMI error to confirm that a hardware/software state mismatch is visible.

PR #28, opened 2026-08-31
rtl837x: clear global private state on teardown

MNeroba:
## Summary

Clear `rtl_gbl_priv` when probe fails after publishing the switch context and when the MDIO device is removed.

## Why this is correct

The Realtek ASIC accessors in `rtk-api/dal/rtl8373/rtl8373_asicdrv.c` dereference `rtl_gbl_priv->map` directly. `rtl837x_dsa_probe()` assigns `rtl_gbl_priv = gsw` before `rtl8372n_hw_init()` and before DSA registration. If either stage fails, the probe path frees `gsw` but previously left the global pointing at that freed object. The normal remove path also completed teardown without clearing it.

The helper uses an identity check before clearing. This keeps a cleanup path from erasing a newer global value if another instance has already published its context. It is called only after all remaining teardown operations that can use the vendor API, and before the failed probe object is freed.

## Scope

This is a lifecycle-only cleanup. It does not change register programming, DSA behavior, or the successful probe path.

## Validation

- `git diff --check`
- Static review of every assignment and use of `rtl_gbl_priv` in the driver and ASIC accessors

Please review the global-state lifetime and verify probe failure/rebind scenarios, especially an `rtl8372n_hw_init()` failure followed by another probe and a normal unbind/rebind cycle. I estimate 93% confidence: the stale-pointer lifetime defect and cleanup ordering are clear from the source, while the exact multi-instance expectations of the vendor API should be confirmed by maintainers.

### DIFF
--- a/package/kernel/rtl837x/src/rtl837x_mdio.c
+++ b/package/kernel/rtl837x/src/rtl837x_mdio.c
@@ -729,6 +729,12 @@ static const struct of_device_id rtk_gsw_match[] = {
 
 MODULE_DEVICE_TABLE(of, rtk_gsw_match);
 
+static void rtl837x_clear_global_priv(struct rtk_gsw *gsw)
+{
+	if (rtl_gbl_priv == gsw)
+		rtl_gbl_priv = NULL;
+}
+
 static int rtl837x_dsa_probe(struct mdio_device *mdiodev)
 {
 	struct device *dev =&mdiodev->dev;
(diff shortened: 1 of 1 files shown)


MNeroba:

This is a static lifecycle correction rather than a hardware feature change. The vendor register-access layer uses the global `rtl_gbl_priv` as its only context and dereferences it without a NULL/lifetime check. The probe publishes that pointer before hardware initialization and DSA registration, while both failure paths release the same `gsw` object. The remove path tears down work, SFP, DSA, debugfs, and the master reference, but previously left the global published.

The patch clears the pointer on each post-publication probe failure and after normal remove. The identity check is deliberate: it prevents a cleanup path from clearing a different context if the global has been replaced. Cleanup is after the last teardown operation that can call the vendor API, so no existing operation needs the pointer after it is cleared.

Confidence: 93%. I am highly confident that the current code can leave stale global state and that this patch removes it without changing the successful path. The remaining point for maintainer confirmation is the vendor layer’s intended single-instance/multi-instance lifetime model. Please test failed probe followed by retry, and unbind/rebind, with KASAN/KFENCE or equivalent diagnostics if available.

PR #29, opened 2026-08-31
rtl837x: handle SerDes swap configuration errors

MNeroba:
## Summary

Check all optional SerDes and PHY swap writes during `rtl8372n_hw_init()` and report failures without making the whole switch probe fatal.

The covered operations are the SDS0/SDS1 RX/TX PN-swap pairs, PHY MDI reverse, and PHY TX polarity swap.

## Why this is correct

The mapper and DAL declare these operations as status-returning SMI/MDIO accesses, so failures are real hardware-access errors rather than fire-and-forget calls. The previous code discarded those statuses, making a degraded board-specific configuration invisible.

These writes occur immediately after hardware reset, where early SMI access is known to be transient on this device. Following the maintainer requested policy, failures are reported with `dev_warn()` and initialization continues. This preserves the existing best-effort probe behavior and prevents a transient optional swap write from becoming `-ENODEV` and permanently taking down the LAN path. Each member of an SDS pair is still attempted, and the warning identifies the lane, direction, and page-specific operation that failed.

Mandatory `rtk_switch_init()`, VLAN, IGMP, and other data-path initialization failures retain their existing fail-fast behavior. This patch does not claim transactional rollback for a pair; it makes a partial configuration visible while preserving the established boot semantics.

## Validation

- `git diff --check`: passed.
- Rebased directly onto the current `flint3-be9300` tip.
- Cross-checked mapper return types and the DAL SMI access paths.

No Flint 3 hardware was available, so warning-level choice and degraded-link behavior remain requested maintainer validation.

Please verify a normal BE9300 boot, an early transient SMI failure, and the resulting SerDes and link behavior.

### DIFF
--- a/package/kernel/rtl837x/src/rtl837x_mdio.c
+++ b/package/kernel/rtl837x/src/rtl837x_mdio.c
@@ -463,16 +463,16 @@ int rtl8372n_hw_init(struct rtk_gsw *gsw, rtl837x_pnswap_cfg_t swap_cfg)
 		ret = gsw->pMapper->rtl8373_sds_regbits_write(
 			0, 0, 0, 0x200, 1);
 		if (ret) {
-			dev_err(gsw->dev,
-				"SDS0 RX PN swap configuration failed, error:%d\n", ret);
-			return -EPERM;
+			dev_warn(gsw->dev,
+				 "SDS0 RX PN swap configuration failed, continuing, error:%d\n",
+				 ret);
 		}
 
 		ret = gsw->pMapper->rtl8373_sds_regbits_write(0, 6, 2, 0x2000, 1);
 		if (ret) {
-			dev_err(gsw->dev,
-				"SDS0 RX PN swap configuration failed, error:%d\n", ret);
-			return -EPERM;
+			dev_warn(gsw->dev,
+				 "SDS0 RX PN swap configuration failed, continuing, error:%d\n",
+				 ret);
 		}
 	}
 
@@ -482,16 +482,16 @@ int rtl8372n_hw_init(struct rtk_gsw *gsw, rtl837x_pnswap_cfg_t swap_cfg)
 		ret = gsw->pMapper->rtl8373_sds_regbits_write(
 			0, 0, 0, 1 << 8, 1);
 		if (ret) {
-			dev_err(gsw->dev,
-				"SDS0 TX PN swap configuration failed, error:%d\n", ret);
-			return -EPERM;
+			dev_warn(gsw->dev,
+				 "SDS0 TX PN swap configuration failed, continuing, error:%d\n",
+				 ret);
 		}
 
 		ret = gsw->pMapper->rtl8373_sds_regbits_write(0, 6, 2, 1 << 14, 1);
 		if (ret) {
-			dev_err(gsw->dev,
-				"SDS0 TX PN swap configuration failed, error:%d\n", ret);
-			return -EPERM;
+			dev_warn(gsw->dev,
+				 "SDS0 TX PN swap configuration failed, continuing, error:%d\n",
+				 ret);
 		}
 	}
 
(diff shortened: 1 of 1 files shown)


MNeroba:

The patch covers ten status-returning hardware writes that are currently unchecked in `rtl8372n_hw_init()`: two writes for each SDS0/SDS1 RX/TX PN-swap option plus the MDI-reverse and TX-polarity-swap register writes. The mapper type definitions and DAL implementations show that these functions can return SMI failures; they are not void fire-and-forget helpers.

Failing the probe after a requested board-level wiring correction cannot make the configuration worse than silently continuing with only part of the correction applied. The patch preserves the existing `-EPERM` convention already used for other hardware-init failures and leaves the successful path unchanged.

Confidence: 95%. Source-level correctness is high because the return types and error propagation are explicit. The requested maintainer check is hardware-specific: confirm that every device-tree option means the full pair/operation is mandatory and that no board intentionally accepts a partially programmed swap configuration. Suggested test: enable each option separately, force or observe an SMI write failure, and verify that probe aborts with the corresponding error log.

perceival:
Thanks — the error checking itself is right, but as written this turns best-effort
SerDes writes into a permanent probe failure, and this board cannot afford that.

Every `rtl8373_sds_regbits_write` / `setAsicRegBits` failure now returns `-EPERM`,
which the caller at `rtl837x_mdio.c:857` turns into `-ENODEV` from probe — not
`-EPROBE_DEFER`, so the switch never comes back without a reboot. Two problems:

1. These are read-modify-write MDIO/SMI accesses issued right after
   `rtl837x_hw_reset()`. `rtl837x_switch_probe()` already retries chip detection four
   times with 50 ms delays precisely because early SMI access on this part is flaky,
   and `dal_rtl8373_sds_regbits_write()` has no init-state gate, so the only
   non-`RT_ERR_INPUT` failure mode here is exactly that transient bus error. Today a
   hiccup is swallowed; with this patch it kills the LAN data path outright.
2. It bails mid-sequence — e.g. after the `page0 reg0` write but before the paired
   `page6 reg2` write — leaving the SerDes half-swapped, with no re-init path.

The GL-BE9300 sets all four `sdsN-rx/tx-swap` flags, so every one of these branches
runs on that board on every boot.

Could you either log and continue (preserving current semantics), or propagate the
real error and have probe return `-EPROBE_DEFER` so the device is retried? Happy to
merge either shape.

MNeroba:
Thanks, agreed. I pushed a follow-up commit (`6cdccadfd9`) that takes the best-effort path you suggested.

All optional SerDes/PHY swap write failures are now reported with `dev_warn()` and initialization continues, so a transient early SMI error no longer becomes `-ENODEV` and permanently prevents the switch from probing. The paired second write is still attempted even if the first write fails, while the mandatory `rtk_switch_init()`, VLAN, IGMP, and other data-path setup failures retain their existing fail-fast handling.

This preserves the current boot semantics and makes a partial swap visible in the kernel log. Please check whether the warning level/message is appropriate for the board; I can rework this to an explicit retry/`-EPROBE_DEFER` path if you prefer that stricter behaviour.

MNeroba:
Rebased onto the current `flint3-be9300` base and pushed as `b5de45de84`; the best-effort behavior requested in review is preserved.

The four SDS swap pairs and the optional PHY polarity/MDI writes now report failures with `dev_warn()` and continue, so a transient early SMI error cannot turn an optional board-specific swap into a permanent `-ENODEV` probe failure. The mandatory `rtk_switch_init()` path remains fail-fast. Each SDS pair still attempts both writes, which avoids silently skipping the paired hardware setting after the first access fails.

Static confidence is about 96% for the intended probe semantics. Hardware validation remains useful for the BE9300's four swap flags, especially a boot with an injected early MDIO/SMI failure and a subsequent link check; that is the point where we can confirm whether a warning-only degraded SerDes configuration is preferable to losing the whole switch, as discussed.

perceival:
**Merging.** Verified the premise: `rtl8373_sds_regbits_write` and `rtl8373_setAsicRegBits` are both genuinely status-returning in the mapper, so those return values really were being discarded. `ret` is already declared in `rtl8372n_hw_init` and is reassigned by `rtk_switch_init()` immediately after the block, so no stale value escapes. Behaviourally this is a pure logging change — the writes were fire-and-forget before and remain non-fatal now, so nothing regresses. Thanks for taking the best-effort route.

Two observations, neither blocking:

- `ret` is `unsigned int` in that function and the new `dev_warn(..., "%d", ret)` prints it with `%d`. Harmless in practice (RT_ERR codes are small positives), just technically mismatched.
- Worth being clear about what the warning does and does not buy us: if the first write of an SDS pair succeeds and the second fails, the lane is left half-swapped. On BE9300 the LAN path is dead without those four swap flags, so the outcome is "dead LAN plus one `dev_warn`". That is the same outcome as today, only visible instead of silent — a real improvement, but it makes the failure *diagnosable*, not *survivable*. Since each block issues two writes that share one identical message string, consider distinguishing them so the log says which register failed.

MNeroba:
Pushed `17aee86e0f` to address the follow-up review. The vendor return value is now signed for `%d` diagnostics, and the best-effort warnings identify the affected SerDes lane, direction, and page. The PR description was updated accordingly.

PR #30, opened 2026-08-31
rtl837x: report CPU tag teardown errors

MNeroba:
## Summary

Report failures from the CPU-tag disable operation in the DSA teardown callback.

## Why this is correct

`rtk_cpuTag_enable_set(EXTERNAL_CPU, DISABLED)` is a status-returning Realtek API. Its documented return values include SMI access failures, invalid input, and driver errors. The setup path already treats the same operation as mandatory and aborts on failure; teardown currently discards the status because the DSA callback itself is `void`.

This patch preserves the `void` callback contract and does not change teardown control flow. It only emits an error when the hardware cannot be left with the proprietary CPU tag disabled, so a failed cleanup is diagnosable instead of silent.

## Validation

- `git diff --check`
- Compared the teardown call with the checked setup call and the Realtek `cpuTag` API declaration/implementation

Please verify whether the CPU-tag disable is required after `rtl837x_mdio_teardown()` on all supported chips and, if possible, exercise unbind/rebind with an injected SMI failure. I estimate 96% confidence: the call returns an explicit error and the DSA callback cannot propagate it, so logging is the least invasive correct handling.

### DIFF
--- a/package/kernel/rtl837x/src/rtl837x_dsa_ops.c
+++ b/package/kernel/rtl837x/src/rtl837x_dsa_ops.c
@@ -342,12 +342,16 @@ static int rtl837x_setup(struct dsa_switch *ds)
 
 static void rtl837x_teardown(struct dsa_switch *ds)
 {
+	int ret;
+
 	rtnl_lock();
 	dsa_tag_8021q_unregister(ds);
 	rtnl_unlock();
 
 	rtl837x_mdio_teardown(ds);
-	rtk_cpuTag_enable_set(EXTERNAL_CPU, DISABLED);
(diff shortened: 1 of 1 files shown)


MNeroba:

The teardown callback is `void`, so it cannot return the `rtk_cpuTag_enable_set()` status to DSA. The patch captures the status and uses `dev_err()` only on failure. This follows the driver’s existing setup behavior, where the same API is checked and a nonzero Realtek status aborts setup.

Confidence: 96%. The change is intentionally diagnostic and has no effect when the operation succeeds. The maintainer-specific check is whether the final disable is still meaningful after MDIO/DSA teardown for every supported chip; an injected SMI failure or a controlled unbind/rebind would validate the log path.

PR #31, opened 2026-08-31
rtl837x: handle debugfs SerDes dump errors

MNeroba:
## Summary

Stop the debugfs SerDes page dump when any underlying register read fails.

## Why this is correct

`_sds_page_dump_read()` performs 21 Realtek register/bit reads and previously ignored every return value. The vendor APIs return status codes for SMI failures and may leave the output argument unchanged when a read fails. The handler then formatted `v3` as if the read had succeeded, so the dump could contain stale or uninitialized data and appear valid.

This patch checks each status before formatting its value. On failure it frees the temporary 4096-byte buffer and returns `-EIO`, which is the appropriate result for a debugfs read backed by a failed hardware transaction. The successful dump content and ordering are unchanged.

## Scope

Only the `sds_page_dump` read path is changed. The separate debugfs write-buffer and command-parser fixes are intentionally left to PRs #21 and #22.

## Validation

- `git diff --check`
- Static review of all 21 calls in the dump handler against the status-returning Realtek API declarations and implementations

Please exercise the dump with a healthy switch and with an injected/observed SMI failure at several read positions, confirming that success output is unchanged and failure returns `-EIO` without a partial misleading dump. I estimate 95% confidence: the ignored status values and cleanup path are directly visible in source; only the preferred user-facing debugfs error presentation needs maintainer confirmation.

### DIFF
--- a/package/kernel/rtl837x/src/rtl837x_dbg.c
+++ b/package/kernel/rtl837x/src/rtl837x_dbg.c
@@ -142,51 +142,99 @@ static ssize_t _sds_page_dump_read(struct file *filep, char __user *ubuf,
 	if (!buf)
 		return -ENOMEM;
 
-	rtk_rtl8373_getAsicReg(RTL8373_SDS_MODE_SEL_ADDR, &v3);
+	ret = rtk_rtl8373_getAsicReg(RTL8373_SDS_MODE_SEL_ADDR, &v3);
+	if (ret)
+		goto out_error;
 	len += snprintf(buf + len, 4096 - len, "reg 0x7b20: %#x\n", v3);
-	rtk_rtl8373_sds_reg_read(0, 0x21, 0x10, &v3);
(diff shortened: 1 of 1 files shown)


MNeroba:

The dump handler reuses one `v3` variable for all hardware reads. Because `rtk_rtl8373_getAsicReg()`, `rtk_rtl8373_sds_reg_read()`, and `rtk_rtl8373_sds_regbits_read()` report SMI failures through their return values, ignoring those values allows the next `snprintf()` to present data that was not obtained by the requested transaction. The patch checks before every formatting step and exits through one cleanup path.

Confidence: 95%. This is a local control-flow/data-validity fix with no change to the successful register sequence. Maintainers should confirm the desired debugfs UX for a failed read; the conservative behavior is to return `-EIO` and avoid returning a partial dump. Suggested test: inject a failure on the first, middle, and final read and verify no stale value is emitted and the allocation is released.

PR #32, opened 2026-08-31
rtl837x: reject malformed extra-init properties

MNeroba:
## Summary

Reject malformed `extra-init` Device Tree properties whose byte length is not an exact multiple of one register/mask/value tuple.

The current code calculates the tuple count with integer division. A property with one or two trailing cells is therefore accepted and the trailing data is silently ignored. A property shorter than one tuple is also treated as a successful no-op.

* The driver consumes `extra-init` as repeated triples of 32-bit big-endian cells: register, mask, and value.
* `of_property_count_elems_of_size()` is the kernel Device Tree helper for requiring a property length to be aligned to a requested element size.
* The property is optional, so the existing absent/empty-property fast path is preserved.
* A valid property produces the same tuple count and the same register writes as before.
* A malformed property now returns the kernel helper error and stops hardware initialization before applying a partial configuration.
* This is independent of the write-error propagation in PR #12: that PR handles failures after a tuple has been decoded, while this patch validates the tuple container itself.

## Validation

* `git diff --check`
* Cross-checked the helper semantics against the Linux OF implementation.

## Requested review

Please confirm that the intended `extra-init` ABI is exactly a sequence of register/mask/value triples and that rejecting a malformed property is preferable to silently ignoring its trailing cells. If this is accepted, please also test one valid tuple and one malformed property with a trailing cell using a test DTB or an equivalent fault-injection path.

## Confidence

Approximately 93% for the parsing and failure-path correction. The valid-property behavior is unchanged and the change uses the standard kernel helper. The remaining point for maintainer confirmation is whether any external DTS/DTB intentionally relies on non-tuple trailing data.

### DIFF
--- a/package/kernel/rtl837x/src/rtl837x_mdio.c
+++ b/package/kernel/rtl837x/src/rtl837x_mdio.c
@@ -10,6 +10,7 @@
 #include <linux/bits.h>
 #include <linux/device.h>
 #include <linux/delay.h>
+#include <linux/of.h>
 #include <linux/of_mdio.h>
 #include <linux/of_gpio.h>
 #include <linux/of_net.h>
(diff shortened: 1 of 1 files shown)


MNeroba:
Follow-up review note

I am approximately 93% confident in this narrow parsing fix. The existing implementation treats extra-init as register/mask/value triples but uses floor division, so malformed Device Tree data can be silently truncated. The standard OF element-count helper rejects a non-aligned property length; absent and empty properties still retain the existing no-op behavior, and valid triples produce the same writes.

This complements PR #12 rather than duplicating it: #12 handles an error returned by a decoded register write, while this change rejects malformed tuple storage before any write is attempted. Please confirm the intended extra-init ABI and test one valid tuple plus a DTB with a trailing cell. No hardware is needed for the parser check, but maintainer confirmation is requested for compatibility with any external DTBs.

PR #34, opened 2026-08-31
rtl837x: roll back VLAN shadow state on write errors

MNeroba:
## Summary

Restore the RTL837x VLAN shadow entry when programming the corresponding hardware VLAN entry fails.

The DSA VLAN callbacks update `gsw->vlan_table[vid]` before calling `rtl837x_write_vlan()`. If `rtk_vlan_set()` fails, the callback returns an error but the shadow entry keeps the uncommitted membership, tagging, and validity changes. A later retry or another VLAN operation can then build its hardware request from state that was never accepted by the switch.

This applies consistently to both the `tag_8021q` VLAN callbacks and the regular DSA `port_vlan` callbacks.

* DSA reports the callback error to switchdev, so a failed VLAN operation must not be represented as committed in the driver shadow state.
* The old entry is copied only after the port and VLAN ID have passed validation.
* On a failed `rtl837x_write_vlan()`, the complete anonymous VLAN entry is restored, including `valid`, `vid`, `mbr`, and `untag`.
* On success, the existing state transition and hardware programming are unchanged.
* PVID updates remain after a successful VLAN write and are intentionally outside this change; this patch addresses the direct VLAN-table write transaction only.

## Validation

* `git diff --check`
* Reviewed all four VLAN add/delete callbacks for matching rollback behavior.
* Cross-checked the DSA/switchdev error-return contract.

## Requested review

Please verify with a forced `rtk_vlan_set()`/SMI write failure that:

1. The callback returns an error.
2. A subsequent retry starts from the pre-failure shadow membership and tagging state.
3. No successful VLAN operation changes behavior.
4. The same behavior is present for both `tag_8021q` and regular DSA VLAN callbacks.

The bridge VLAN functional behavior is not intentionally changed; this is an error-path consistency fix.

## Confidence

Approximately 94%. The issue is a direct state-ordering mismatch visible in the callback code, and the rollback is limited to the exact entry whose hardware write failed.

### DIFF
--- a/package/kernel/rtl837x/src/rtl837x_dsa_ops.c
+++ b/package/kernel/rtl837x/src/rtl837x_dsa_ops.c
@@ -201,10 +201,13 @@ static int rtl837x_tag_8021q_vlan_add(struct dsa_switch *ds, int port, u16 vid,
 	struct rtk_gsw *gsw = ds->priv;
 	bool untagged = flags & BRIDGE_VLAN_INFO_UNTAGGED;
 	bool pvid = flags & BRIDGE_VLAN_INFO_PVID;
+	typeof(gsw->vlan_table[0]) old_vlan;
 	int ret;
 
 	if (!rtl837x_valid_port(gsw, port) || !vid || vid > RTK_VID_MAX)
 		return -EINVAL;
+
+	old_vlan = gsw->vlan_table[vid];
 
 	gsw->vlan_table[vid].valid = 1;
 	gsw->vlan_table[vid].vid = vid;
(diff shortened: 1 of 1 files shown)


MNeroba:
Follow-up review note

I am approximately 94% confident in this narrow consistency fix. All four VLAN add/delete callbacks modify gsw->vlan_table before calling rtl837x_write_vlan(). When that hardware write fails, the callback correctly returns an error but previously left the uncommitted shadow entry behind. The new code restores the exact prior entry, including validity, membership, and tagging, while leaving successful programming unchanged.

Please fault-inject an SMI/VLAN-table write failure and retry the same operation. The retry should see the pre-failure shadow state. Please also confirm that both the regular DSA VLAN path and the tag_8021q path behave identically.

MNeroba:
Superseded by consolidated PR #42. VLAN-table shadow rollback is retained together with the corresponding PVID rollback so the DSA VLAN state-consistency invariant is reviewed as one unit.

PR #37, opened 2026-08-31
rtl837x: roll back PVID shadow state on write errors

MNeroba:
## Summary

Roll back the driver PVID shadow state when programming the hardware PVID fails.

The `tag_8021q_vlan_add/del` and regular `port_vlan_add/del` callbacks update their `tag8021q_pvid` or `bridge_pvid` fields before calling `rtl837x_commit_pvid()`. If `rtk_vlan_portPvid_set()` fails, the callback returns an error but the shadow state still describes the rejected PVID operation. Later filtering or VLAN operations can then use a PVID that was never committed to the switch.

* The DSA callback returns the hardware error, so the associated software shadow must remain at the pre-operation value.
* Each path saves both the previous VID and its validity flag before changing the shadow state.
* On a failed PVID write, the saved pair is restored. On success, the existing state and `port_pvid` update behavior are unchanged.
* The fix covers both internal `tag_8021q` VLAN management and regular bridge VLAN management.
* This is complementary to PR #34: #34 rolls back the VLAN membership entry when the VLAN-table write fails, while this patch handles the subsequent PVID write transaction.

## Validation

* `git diff --check`
* Reviewed add and delete paths for both PVID shadow arrays.
* Cross-checked the existing `rtl837x_commit_pvid()` ordering: `port_pvid` is updated only after a successful hardware write.

## Requested review

Please fault-inject a failure from `rtk_vlan_portPvid_set()` during:

1. A tag_8021q PVID add.
2. A tag_8021q PVID delete.
3. A bridge PVID add.
4. A bridge PVID delete.

In each case, the DSA callback should return an error and a subsequent retry should start from the pre-failure shadow state. Please also confirm that successful PVID changes remain unchanged.

## Confidence

Approximately 94%. The issue is a direct ordering mismatch between the returned hardware error and the persistent shadow update; the rollback is limited to the exact state changed by the failed operation.

### DIFF
--- a/package/kernel/rtl837x/src/rtl837x_dsa_ops.c
+++ b/package/kernel/rtl837x/src/rtl837x_dsa_ops.c
@@ -220,9 +220,17 @@ static int rtl837x_tag_8021q_vlan_add(struct dsa_switch *ds, int port, u16 vid,
 		return ret;
 
 	if (pvid) {
+		u16 old_pvid = gsw->tag8021q_pvid[port];
+		bool old_pvid_valid = gsw->tag8021q_pvid_valid[port];
+
 		gsw->tag8021q_pvid[port] = vid;
 		gsw->tag8021q_pvid_valid[port] = true;
-		return rtl837x_commit_pvid(gsw, port);
+		ret = rtl837x_commit_pvid(gsw, port);
+		if (ret) {
+			gsw->tag8021q_pvid[port] = old_pvid;
+			gsw->tag8021q_pvid_valid[port] = old_pvid_valid;
+		}
+		return ret;
 	}
 
 	return 0;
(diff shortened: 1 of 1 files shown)


MNeroba:
Follow-up review note

I am approximately 94% confident in this state-consistency fix. The callbacks set tag8021q_pvid or bridge_pvid before rtl837x_commit_pvid(), but commit_pvid updates the hardware first and only updates port_pvid after success. If the hardware PVID write fails, the callback returns an error while the selected shadow pair remains changed; the new rollback restores the previous VID and validity flag.

Please fault-inject rtk_vlan_portPvid_set() for add and delete in both the tag_8021q and bridge paths, then retry the operation. The retry should start from the pre-failure shadow state, while successful operations should remain identical.

MNeroba:
Superseded by consolidated PR #42. PVID shadow rollback is retained together with VLAN-table rollback so the DSA VLAN state-consistency invariant is reviewed as one unit.

PR #38, opened 2026-08-31
rtl837x: avoid pinning status work to current CPU

MNeroba:
## Summary

The periodic CPU-port status worker queued its delayed work with `queue_delayed_work_on(smp_processor_id(), system_wq, ...)` in both the initial setup and the self-requeue path.

This changes both call sites to `queue_delayed_work(system_wq, ...)`.

## Why this is the correct API

`rtl837x_status_check_work_init()` runs from probe context and `rtl837x_status_check_work_func()` runs from a preemptible workqueue context. Neither caller establishes a stable CPU context before calling `smp_processor_id()`. Linux documents that `smp_processor_id()` requires IRQs or preemption to be disabled, or a CPU-affine task, and may warn with `CONFIG_DEBUG_PREEMPT`:

- https://github.com/torvalds/linux/blob/master/include/linux/smp.h#L243-L262

The generic delayed-work helper uses `WORK_CPU_UNBOUND`, which means the workqueue may prefer the local CPU without requiring the caller to read an unstable CPU number:

- https://github.com/torvalds/linux/blob/master/include/linux/workqueue.h#L658-L670

The worker has no per-CPU state: it operates on the shared `struct rtk_gsw` instance and the switch API. Removing the explicit CPU argument therefore avoids an invalid CPU-ID assumption and an unnecessary hard pin while retaining the existing system workqueue and delay.

The legacy `rtl837x_swconfig.c` source is not part of the active `rtl837x_dsa` object list, so this PR is limited to the compiled DSA path.

## Validation

- `git diff --check`
- Full target build and hardware testing were not available in this environment.

## Confidence and requested review

Static confidence: 92%. Please confirm that status polling has no intentional CPU-affinity requirement. On hardware or with `CONFIG_DEBUG_PREEMPT`, please verify that the worker continues to poll CPU-port link state and perform the existing SerDes recovery without warnings or behavior changes.

### DIFF
--- a/package/kernel/rtl837x/src/rtl837x_mdio.c
+++ b/package/kernel/rtl837x/src/rtl837x_mdio.c
@@ -567,11 +567,8 @@ static void rtl837x_status_check_work_func(struct work_struct *work)
 		mdelay(2000);
 	}
 
-	queue_delayed_work_on(smp_processor_id(), 
-						system_wq, 
-						&gsw->status_check_work, 
-						msecs_to_jiffies(gsw->default_work_delay_ms)
-					);
+	queue_delayed_work(system_wq, &gsw->status_check_work,
+				   msecs_to_jiffies(gsw->default_work_delay_ms));
 }
 
 /* unused */
@@ -674,11 +671,8 @@ static int rtl837x_status_check_work_init(struct rtk_gsw *gsw)
 {
 	gsw->default_work_delay_ms = 1000;
 	INIT_DELAYED_WORK(&gsw->status_check_work, rtl837x_status_check_work_func);
-	queue_delayed_work_on(smp_processor_id(), 
-						system_wq, 
-						&gsw->status_check_work, 
-						msecs_to_jiffies(gsw->default_work_delay_ms)
-					);
+	queue_delayed_work(system_wq, &gsw->status_check_work,
+				   msecs_to_jiffies(gsw->default_work_delay_ms));
 	return 0;
 }
 


MNeroba:

I am approximately **92% confident** this is the correct change based on the Linux workqueue and SMP APIs.

Both active call sites obtain a CPU number with `smp_processor_id()` from preemptible context. The kernel documents that this accessor is only valid when preemption or interrupts are disabled, or when the task is CPU-affine; otherwise `CONFIG_DEBUG_PREEMPT` can report a warning. `queue_delayed_work()` is the purpose-built helper for this case: it keeps the existing system per-CPU workqueue and delay, while letting the workqueue use the local CPU without exposing an unstable CPU-ID read.

The status callback has no per-CPU data or CPU-local register state. Its state is the shared `struct rtk_gsw`, and the RTK switch accesses are already serialized by the driver/API layers. Therefore moving from an explicit CPU argument to the generic helper should not change the switch behavior.

Please specifically verify one point before merging: that no undocumented CPU-affinity requirement exists for the MDIO/SerDes status worker. If affinity is intentional, the safer alternative would be to preserve the pinning while bracketing the CPU lookup with the appropriate preemption protection. Otherwise, the submitted change removes the warning-prone assumption and matches the kernel helper semantics.

Suggested checks:

- build with `CONFIG_DEBUG_PREEMPT=y` and confirm there is no `smp_processor_id()` warning;
- confirm the periodic worker still polls CPU-port status and performs the existing SerDes recovery;
- inspect behavior on CPU hotplug or any target setup that changes CPU availability.

MNeroba:
Superseded by consolidated PR #41. The scheduling change is retained with the rest of the status-worker lifecycle so its CPU-affinity assumptions can be checked together.

PR #40, opened 2026-08-31
rtl837x: handle invalid SerDes mode properties

MNeroba:
## Summary

Handle invalid RTL837x SerDes mode properties without making the entire switch
probe fail.

- An absent optional property keeps the existing `SERDES_OFF` default.
- Malformed, empty, and unknown values emit a warning and leave the affected
  SerDes lane disabled.
- The parser has no deferred dependency; property errors are local and never
  abort the probe.
- Valid values retain the existing SerDes mode selection.

## Why this is correct

Device-tree input errors are isolated to the affected lane. This parser reads
only local string properties and has no `-EPROBE_DEFER` path, so malformed or
unknown configuration should not be converted into a probe retry. This keeps
the failure policy explicit and avoids silently enabling an unknown SerDes
configuration.

## Validation

- Existing source review confirms the behavior above.
- Full target build and hardware validation remain to be run on a Linux/OpenWrt build host.

### DIFF
--- a/package/kernel/rtl837x/src/rtl837x_mdio.c
+++ b/package/kernel/rtl837x/src/rtl837x_mdio.c
@@ -368,7 +368,22 @@ static int rtl837x_sdsmode(const char *name, rtk_sds_mode_t *mode)
 		}
 	}
 
-	return -1;
+	return -EINVAL;
+}
+
+static int rtl837x_parse_sdsmode(struct device_node *np, const char *propname,
+				 rtk_sds_mode_t *mode)
+{
+	const char *name;
+	int ret;
+
+	ret = of_property_read_string(np, propname, &name);
+	if (ret == -EINVAL)
+		return 0;
+	if (ret)
+		return ret;
+
+	return rtl837x_sdsmode(name, mode);
 }
 
 static int rtl8372n_igmp_init(struct rtk_gsw *gsw)
@@ -817,8 +832,6 @@ static int rtl837x_dsa_probe(struct mdio_device *mdiodev)
 	struct rtk_gsw *gsw;
 	struct device_node *ethernet;
 	struct net_device *master;
-	const char *sdsmode_name;
-	rtk_sds_mode_t sdsmode;
 	struct regmap_config rc;
 	u32 cpu_port;
 	bool cpu_port_from_dsa = false;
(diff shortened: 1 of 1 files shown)


MNeroba:

I am approximately **94% confident** in this change because the failure is visible from the parser control flow, without requiring hardware: the previous compound condition silently ignored both a malformed DT string and a string that was not present in the driver mode table.

A missing property still leaves the existing `SERDES_OFF` default untouched. For a present property, however, silently falling back to `SERDES_OFF` is unsafe for `sds0mode`, which is used for the CPU SerDes path on Flint 3. Probe can otherwise succeed while the required link remains disabled. Returning the OF error or `-EINVAL` makes the configuration failure explicit before switch initialization and DSA registration.

This follows the same boundary used by Linux `of_get_phy_mode()`: an optional property may be absent, but an invalid value that is present is not silently accepted. The supported mode strings remain exactly the local table entries; no new hardware mode is being invented.

Please confirm two assumptions before merging:

- both SerDes properties are optional when absent, with `SERDES_OFF` as the intended default;
- no downstream DTS uses a legacy alias or an intentionally ignored mode string.

Suggested checks are the current Flint 3 DT, an omitted `sds1mode`, a misspelled mode, and malformed/empty property data. The last two should fail before any global driver state or DSA registration is established.

perceival:
Thanks for the rework — 39, 41, 42, 44 and 45 look good and I'm merging them. This
one still has the pattern I flagged on #29/#18/#19, though.

`rtl837x_parse_sdsmode()` now returns `-EINVAL` for an unrecognised mode string, and
both call sites do `dev_put(master); return ret;`. So a typo'd — or simply
future-named — `rtl837x,sdsNmode` value in DT kills the entire DSA switch, rather than
falling back to the `SERDES_OFF` default already set just above. It also aborts on
`-ENODATA`/`-EILSEQ`, i.e. a property that is present but empty.

Keeping `rtl837x_sdsmode()` returning `-EINVAL` is right; it's the call site that
should be forgiving:

```c
if (ret == -EPROBE_DEFER) { dev_put(master); return ret; }
if (ret) dev_warn(dev, "bad sds%d mode property, leaving SerDes off: %d\n", i, ret);
```

Also: the `#include <linux/of.h>` hunk duplicates an include that is already at tip, so
it will conflict on rebase.

All of these branches are cut from a stale base — please rebase onto `7eed649fd8`.

MNeroba:
## Follow-up after review

I re-read the latest maintainer comment before updating this PR. No newer author
commit was present. Commit 17072152e8d5e90b144c8988e6a96de7efb4029f incorporates
the requested behavior:

- an absent property keeps the existing SERDES_OFF default;
- malformed, empty, unknown, and future values are reported with dev_warn and
  leave the affected lane disabled;
- only EPROBE_DEFER aborts probe and requests a retry;
- the duplicate linux/of.h include was not carried over because it is already
  present at the current base.

This keeps the parser strict while preventing an optional or future DT value
from taking down the complete DSA switch. It also preserves the exact parser
error for diagnostics and does not invent a new SerDes mode.

Validation passed with git diff --check and checkpatch reported no code errors.
The only checkpatch warning concerns line wrapping in the commit description.
A full target build and hardware test are not available in this environment.

Please verify the current Flint 3 DT, an omitted sds1mode, a misspelled mode,
and empty or wrong-type property data. Static confidence is approximately 98
percent; the remaining uncertainty is limited to downstream DTS assumptions
and runtime confirmation.

perceival:
**Merging with one request.** The API reasoning is correct and I verified it against `drivers/of/property.c`: `of_property_read_string()` returns exactly `-EINVAL` (absent), `-ENODATA` (zero length), `-EILSEQ` (not NUL-terminated), or 0. So mapping `-EINVAL` to "absent, keep `SERDES_OFF`" is right and cannot be confused with an unknown-string value, since unknown strings only reach `rtl837x_sdsmode()` after a successful read. Passing `&gsw->sds0mode` directly is safe — `rtl837x_sdsmode()` writes `*mode` only on a match and both fields are pre-set to `SERDES_OFF`. Both old callers are replaced and nothing else consumes the `-1` -> `-EINVAL` change.

**Request: please drop the `-EPROBE_DEFER` branches.** They are unreachable. `rtl837x_parse_sdsmode()` can only ever return 0, `-EINVAL`, `-ENODATA` or `-EILSEQ` — a string-property parse has no provider dependency to defer on. That makes both blocks (~24 lines) dead code, and the summary's "Only `-EPROBE_DEFER` aborts the probe" describes a path that cannot execute. It also implies a dependency the property doesn't have, which is misleading to the next reader.

(Contrast with #43, where the same pattern is correct — `devm_gpiod_get_optional()` really can defer.)

Note this and #43 both touch the reset-GPIO block, so whichever lands second will need a trivial rebase.

MNeroba:
Pushed `112c32318c` to remove the unreachable `-EPROBE_DEFER` branches from local SerDes property parsing. Missing properties retain defaults; malformed, empty, and unknown values warn and disable the affected SerDes mode. The body now reflects that property parsing cannot defer probe.

PR #42, opened 2026-08-31
rtl837x: keep VLAN shadow state consistent on errors

MNeroba:
## Summary

Consolidate PRs #34 and #37 into one transactional shadow-state fix.

The DSA VLAN callbacks now restore:

- VLAN membership/tagging shadow state when the VLAN-table write fails;
- tag_8021q and bridge PVID shadow state when the PVID write fails.

Both regular bridge VLAN and internal tag_8021q paths are covered for add and delete operations.

The driver updates its software shadow before programming the corresponding hardware state. If the hardware operation returns an error, retaining the new shadow would make later operations trust a change that was not accepted by the switch. The combined change restores exactly the fields modified by the failed operation and leaves successful operations unchanged.

## Review organization

During the audit of PRs #10-#40, these two changes were found to implement the same invariant across the same four VLAN callback pairs. They were consolidated into one review unit for easier verification and confirmation. This does not add a new hardware feature or change the successful programming path.

## Validation

- git diff --check

## Review request

Please fault-inject VLAN-table and PVID write failures in both regular DSA VLAN and tag_8021q paths, for add and delete. A failed operation should return an error and the next retry should start from the pre-operation shadow state.

## Confidence

Approximately 95% for the state-consistency logic. Runtime confirmation is requested for partial hardware-failure behavior and retries.

### DIFF
--- a/package/kernel/rtl837x/src/rtl837x_dsa_ops.c
+++ b/package/kernel/rtl837x/src/rtl837x_dsa_ops.c
@@ -201,10 +201,13 @@ static int rtl837x_tag_8021q_vlan_add(struct dsa_switch *ds, int port, u16 vid,
 	struct rtk_gsw *gsw = ds->priv;
 	bool untagged = flags & BRIDGE_VLAN_INFO_UNTAGGED;
 	bool pvid = flags & BRIDGE_VLAN_INFO_PVID;
+	typeof(gsw->vlan_table[0]) old_vlan;
 	int ret;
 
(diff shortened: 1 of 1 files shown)


MNeroba:
Review-organization note: after auditing PRs #10-#40, the related VLAN-table and PVID shadow rollback changes from #34 and #37 were consolidated here. They implement one state-consistency invariant across the same DSA VLAN paths, which should make review and failure-injection confirmation easier.

PR #43, opened 2026-08-31
rtl837x: handle reset GPIO lifecycle correctly

MNeroba:
## Summary

Consolidate the reset-GPIO lifecycle fix while keeping the GPIO optional.

- use IS_ERR_OR_NULL before attempting a reset;
- use gpiod_set_value_cansleep for providers that may sleep;
- abort probe only for EPROBE_DEFER;
- warn and clear the descriptor for other optional-GPIO errors so the switch
  continues without reset control.

The reset levels, polarity handling, and delays are unchanged for a valid
descriptor.

## Why this is correct

The optional GPIO consumer API distinguishes an absent descriptor from an
ERR_PTR acquisition failure. A reset GPIO is auxiliary to the data path, so a
non-deferred acquisition problem must not unregister the DSA switch. A deferred
provider still requires the normal probe retry. The reset sequence runs in
probe context and therefore must use the sleepable accessor. This path is
before drvdata and global private state are installed, so no stale global
pointer is introduced by the defer path.

## Validation

- git diff --check: passed;
- checkpatch: no code errors; the only warning is commit-message line wrapping;
- full target build and Flint 3 hardware testing are not available in this
  checkout.

## Review request

Please verify device trees with no reset GPIO, a normal GPIO, a sleep-capable
GPIO provider, an acquisition failure, and EPROBE_DEFER. The normal reset
waveform should remain unchanged. Static confidence is approximately 98
percent; runtime confirmation is requested for the GPIO-provider variants.

### DIFF
--- a/package/kernel/rtl837x/src/rtl837x_mdio.c
+++ b/package/kernel/rtl837x/src/rtl837x_mdio.c
@@ -329,13 +329,13 @@ END_DETECT_CHIP:
 
 static int rtl837x_hw_reset(struct rtk_gsw *gsw)
 {
-	if (!IS_ERR(gsw->reset_pin)) {
+	if (!IS_ERR_OR_NULL(gsw->reset_pin)) {
 		dev_info(gsw->dev, "START HW RESET");
-		gpiod_set_value(gsw->reset_pin, 1);
+		gpiod_set_value_cansleep(gsw->reset_pin, 1);
 		mdelay(100);
-		gpiod_set_value(gsw->reset_pin, 0);
+		gpiod_set_value_cansleep(gsw->reset_pin, 0);
 		mdelay(100);
-		gpiod_set_value(gsw->reset_pin, 1);
+		gpiod_set_value_cansleep(gsw->reset_pin, 1);
 		mdelay(100);
 		dev_info(gsw->dev, "FINISH HW RESET");
 	}
@@ -897,7 +897,18 @@ static int rtl837x_dsa_probe(struct mdio_device *mdiodev)
 
 	gsw->reset_pin = devm_gpiod_get_optional(dev, "reset", GPIOD_OUT_LOW);
 	if (IS_ERR(gsw->reset_pin)) {
-		dev_warn(dev, "failed to get RESET GPIO!!!\n");
+		ret = PTR_ERR(gsw->reset_pin);
+		if (ret == -EPROBE_DEFER) {
+			dev_err_probe(dev, ret, "failed to get reset GPIO\n");
+			if (master)
+				dev_put(master);
+			return ret;
+		}
+
+		dev_warn(dev,
+			 "failed to get reset GPIO: %d; "
+			 "continuing without reset GPIO\n", ret);
+		gsw->reset_pin = NULL;
 	}
 
 	if (!of_property_read_string(np, "rtl837x,sds0mode", &sdsmode_name) &&


MNeroba:
Review-organization note: after auditing PRs #10-#40, the reset descriptor acquisition and reset accessor changes from #17 and #35 were consolidated here. They cover one reset-GPIO lifecycle and are kept together so optional, error, and sleep-capable cases can be checked as one unit.

perceival:
Two of the three changes here are right and I want them: `IS_ERR_OR_NULL()` in
`rtl837x_hw_reset()` (today a NULL from `devm_gpiod_get_optional()` falls into the reset
path, harmless only because `gpiod_set_value()` short-circuits on NULL), and
`gpiod_set_value_cansleep()` — this is probe context and the current call would
`WARN_ON` a sleeping GPIO controller.

The blocker is the new unconditional `return ret;`: it aborts probe for *every*
`devm_gpiod_get_optional()` error, not just deferral. A board with a quirky reset-GPIO
description loses its whole LAN switch over an optional pin. Suggested split:

```c
if (IS_ERR(gsw->reset_pin)) {
        ret = PTR_ERR(gsw->reset_pin);
        if (ret == -EPROBE_DEFER) { dev_put(master); return ret; }
        dev_warn(dev, "reset GPIO unusable (%d), continuing without it\n", ret);
        gsw->reset_pin = NULL;
}
```

Credit where due: this path runs before `dev_set_drvdata()` and `rtl_gbl_priv` are set,
and there's no `devm_kfree()`, so the dangling-global half of my earlier feedback is
properly respected here.

Please rebase onto `7eed649fd8` — these branches are cut from a stale base.

MNeroba:
## Follow-up after review

I re-read the latest maintainer comment before updating this PR. No newer author
commit was present. Commit d12882ec2080c2ecb294352c7fb74046ded2032f keeps both
accepted improvements and applies the requested error policy:

- IS_ERR_OR_NULL protects the reset path when the optional descriptor is absent;
- gpiod_set_value_cansleep is used for sleep-capable GPIO controllers;
- only EPROBE_DEFER aborts probe;
- other optional-GPIO errors are warned about and the descriptor is cleared to
  NULL, so the DSA data path continues without reset control.

The reset levels, polarity handling, and delays remain unchanged for a valid
descriptor. The defer path is before drvdata and global private state are
installed, so it does not introduce the stale-pointer problem discussed in
earlier PRs.

Validation passed with git diff --check and checkpatch reported no code errors.
The only checkpatch warning concerns line wrapping in the commit description.
A full target build and hardware test are not available in this environment.

Please verify device trees with no reset GPIO, a normal GPIO, a sleep-capable
provider, an acquisition failure, and EPROBE_DEFER. Static confidence is
approximately 98 percent; runtime confirmation is requested for the GPIO
provider variants.

perceival:
**Merging.** All three changes check out, and the third is a real bug fix.

- `IS_ERR` -> `IS_ERR_OR_NULL`: honest, though cosmetic in practice. `devm_gpiod_get_optional()` does return `NULL` when the property is absent, but `gpiod_set_value()` on a NULL desc is a silent no-op (`validate_desc()` returns 0 for NULL), so the old code did not actually misbehave. The real gain is no longer printing "START/FINISH HW RESET" for a reset that never happened.
- `gpiod_set_value_cansleep()`: correct. `rtl837x_hw_reset()` is reached only from `rtl837x_dsa_probe()`, which is process context and already `mdelay()`s 300 ms; there is no atomic caller.
- **`-EPROBE_DEFER` propagation: this is the valuable part, and unlike the same pattern in #40 it is genuinely reachable** — `devm_gpiod_get_optional()` returns `-EPROBE_DEFER` when the TLMM provider has not yet registered. Pre-patch, a deferral only printed a warning, kept the ERR_PTR, skipped the reset and carried on probing, so the switch came up having never been reset and `rtl837x_switch_probe()` read whatever the bootloader left behind. Deferring and retrying is right.

Error-path ordering is correct: `dev_put(master)` before returning, and this is before `dev_set_drvdata()`/`rtl_gbl_priv` are installed, so there is no stale global. Setting `gsw->reset_pin = NULL` after a non-defer error is required for the new guard to be correct, and it is there.

Nit, out of scope: the 300 ms of `mdelay()` is now provably in sleepable context and could be `msleep()`.

Heads-up: this and #40 both touch this block, so the second to land needs a rebase.

PR #44, opened 2026-08-31
rtl837x: propagate auxiliary probe failures

MNeroba:
## Summary

Consolidate PRs #18 and #19 into one probe-lifecycle fix for optional auxiliary subsystems.

After the DSA switch has been registered:

- GPIO registration failures are reported with dev_err_probe() and treated as best effort, so an optional GPIO provider cannot take down the LAN data path;
- a failed SFP upstream registration clears gsw->sfp_bus, because the SFP core has already dropped its reference, and then continues without tearing down the DSA switch;
- successful and absent-optional-subsystem paths remain unchanged;
- the optional failure paths do not devm_kfree() gsw or unregister DSA while drvdata and the global private pointer still reference it.

GPIO and SFP are auxiliary features on this board, not prerequisites for the switch data path. A GPIO base collision, gpiolib resource failure, or optional SFP subsystem error should therefore be visible in the log without making the already registered switch unusable.

The SFP pointer must be cleared after sfp_bus_add_upstream() fails: the failed helper has already dropped its own reference, and retaining the pointer would make later removal call sfp_bus_del_upstream() on a bus that was never added. Keeping the main probe object alive until normal devres cleanup also avoids introducing a dangling drvdata or rtl_gbl_priv reference.

## Review organization

During the audit of PRs #10-#40, #18 and #19 were identified as the same post-registration lifecycle class: optional auxiliary setup must report its failure without unwinding the working DSA switch, while failed SFP ownership must be cleared. They were consolidated into one review unit for easier verification and confirmation. The scope is limited to those two paths.

## Validation

- git diff --check

## Review request

Please verify no-GPIO and no-SFP configurations, successful registration, deferred or invalid dependencies, forced GPIO registration failure, forced SFP upstream-registration failure, and remove/unbind cleanup. Confirm that the DSA switch, drvdata, global private pointer, and Ethernet-master reference are released exactly once.

## Confidence

Approximately 95% for the probe control flow and ownership reasoning. Runtime confirmation is requested for the optional-failure and deferred paths.

### DIFF
--- a/package/kernel/rtl837x/src/rtl837x_mdio.c
+++ b/package/kernel/rtl837x/src/rtl837x_mdio.c
@@ -665,6 +665,8 @@ static int rtl837x_sfp_probe(struct rtk_gsw *gsw)
 	gsw->sfp_bus = bus;
 
 	ret = sfp_bus_add_upstream(bus, gsw, &sfp_ops);
+	if (ret)
+		gsw->sfp_bus = NULL;
 	sfp_bus_put(bus);
 
 	return ret;
@@ -875,11 +877,29 @@ static int rtl837x_dsa_probe(struct mdio_device *mdiodev)
 	}
 
 #ifdef CONFIG_GPIOLIB
-	if (of_property_read_bool(np, "gpio-controller")) 
-		rtl837x_gpiochip_init(gsw);
+	if (of_property_read_bool(np, "gpio-controller")) {
+		ret = rtl837x_gpiochip_init(gsw);
+		if (ret) {
+			dev_err_probe(dev, ret,
+				      "failed to register GPIO controller\n");
+			rtl837x_dsa_unregister(gsw);
+			if (master)
+				dev_put(master);
+			devm_kfree(dev, gsw);
+			return ret;
+		}
+	}
 #endif /* CONFIG_GPIOLIB */
 
-	rtl837x_sfp_probe(gsw);
+	ret = rtl837x_sfp_probe(gsw);
+	if (ret) {
+		dev_err_probe(dev, ret, "failed to attach SFP bus\n");
+		rtl837x_dsa_unregister(gsw);
+		if (master)
+			dev_put(master);
+		devm_kfree(dev, gsw);
+		return ret;
+	}
 
 	rtl837x_debug_proc_init(gsw);
 	rtl837x_status_check_work_init(gsw);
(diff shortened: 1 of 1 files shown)


MNeroba:
Review-organization note: after auditing PRs #10-#40, the GPIO-controller and SFP registration failure paths from #18 and #19 were consolidated here. Both use the same post-DSA-registration unwind boundary, so the probe success and cleanup contract can be verified in one review unit.

MNeroba:
Follow-up to the SFP review from #19: I pushed `a623c1d670` and adjusted the consolidated PR.

The confirmed lifetime fix is retained: `rtl837x_sfp_probe()` clears `gsw->sfp_bus` when `sfp_bus_add_upstream()` fails. The optional SFP probe result is no longer propagated from `rtl837x_dsa_probe()`, so an optional-cage failure cannot unregister the already-registered DSA switch or free `gsw` while `drvdata`/`rtl_gbl_priv` still reference it. The GPIO-controller failure path remains fail-fast because that auxiliary device was explicitly requested and its registration contract is different.

This keeps the PR focused on safe cleanup and preserves the existing optional-SFP behavior. Please check the no-SFP, valid-SFP, and registration-failure paths when reviewing.

MNeroba:
Follow-up to the GPIO review from #18: pushed `c897241657`.

`rtl837x_gpiochip_init()` is now reported with `dev_err_probe()` but its failure is non-fatal: the switch continues probing and remains usable without the optional GPIO controller. The new path no longer unregisters DSA, frees `gsw`, or tears down the LAN data path. The SFP result is likewise kept best effort, while the confirmed `gsw->sfp_bus = NULL` cleanup after a failed `sfp_bus_add_upstream()` remains in the earlier follow-up.

This preserves the existing behavior for base collisions/resource exhaustion and keeps the consolidated PR focused on safe auxiliary cleanup. Please re-check GPIO-present/no-GPIO, registration failure, and normal DSA/LAN startup paths.

MNeroba:
Follow-up to the concern raised in closed PR #18: the consolidated PR now uses the requested best-effort shape.

GPIO registration errors are logged with dev_err_probe() and do not tear down the registered DSA switch. A failed optional SFP upstream registration only clears gsw->sfp_bus and continues; it does not unregister DSA, free gsw, or return the auxiliary error from probe. The PR description has been updated to match this behavior.

This preserves the LAN data path when an optional subsystem is unavailable while still avoiding a later SFP remove on a bus that was never added. Please review the failure and remove paths on the BE9300 when convenient.

PR #45, opened 2026-08-31
rtl837x: harden debugfs write handlers

MNeroba:
## Summary

Consolidate PRs #21 and #22 into one hardening change for the three debugfs register-write handlers.

The handlers now:

- require the exact number of parsed fields for each command;
- use unsigned conversions for the unsigned sds_id and port values;
- release the memdup_user_nul() buffer on normal and validation-error paths.

Valid command forms and register operations remain unchanged.

sscanf() returns the number of successful assignments, so checking only for -1 allowed partial input to reach register access code with uninitialized fields. Requiring the expected conversion count prevents that. Since the same validation paths return early, buffer lifetime must be reviewed at the same time; every post-allocation exit now releases the buffer.

## Review organization

While auditing PRs #10-#40, these changes were found to modify the same three write handlers and to interact directly at their early-return paths. They were therefore optimized into one review unit for easier verification and confirmation. This is a review-oriented consolidation; no new debugfs command or hardware operation is introduced.

## Validation

- git diff --check

## Review request

Please test valid read/write commands, empty and partial commands, out-of-range sds_id and port values, repeated writes, and the intended behavior for extra trailing fields.

## Confidence

Approximately 97% for the parser and allocation-lifetime changes. Runtime confirmation is requested for the debugfs interface and repeated-write behavior.

### DIFF
--- a/package/kernel/rtl837x/src/rtl837x_dbg.c
+++ b/package/kernel/rtl837x/src/rtl837x_dbg.c
@@ -49,23 +49,28 @@ ssize_t _sdsreg_rw_write(struct file *filep, const char __user *ubuf,
 		return PTR_ERR(buf);
 	
 	if(buf[0] == 'w') {
-		if(sscanf(buf, "w %d %x %x %x", &sds_id, &page, &reg, &val) == -1)
+		if (sscanf(buf, "w %u %x %x %x", &sds_id, &page, &reg, &val) != 4) {
+			kfree(buf);
 			return -EFAULT;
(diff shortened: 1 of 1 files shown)


MNeroba:
Review-organization note: after auditing PRs #10-#40, the parser validation and allocation-lifetime fixes from #21 and #22 were consolidated here. They modify the same three debugfs write handlers and interact at early-return paths, so reviewing them together makes the complete safety property easier to confirm.

PR #46, opened 2026-08-31
rtl837x: add DSA link aggregation support

MNeroba:
## Summary

Add DSA link aggregation callbacks for the RTL8372/RTL8372N/RTL8373 family.

## Why this implementation is expected to be correct

The bundled RTL8373 RTK API exposes four trunk groups, a four-port-per-group limit, member-port programming, and the same L2/L3/L4 hash inputs used by the DSA LAG hash types. The driver maps `NETDEV_LAG_HASH_L2`, `NETDEV_LAG_HASH_L23`, and `NETDEV_LAG_HASH_L34` directly to the documented `rtk_trunk_distributionAlgorithm_set()` masks and programs members through `rtk_trunk_port_set()` rather than writing undocumented registers.

The DSA switch advertises the four hardware group IDs. CPU-port aggregation is rejected because the current RTL837x unicast FDB API represents one destination port only; this patch deliberately does not add an unverified `lag_fdb_add/del` implementation. A separate feature lock protects the software state and serialized RTK API operations.

## Scope and limitations

This patch provides static hash-based trunk membership. It does not claim that the switch negotiates LACP itself or dynamically removes link-down members; that is intentionally a separate follow-up. Unsupported hash modes, CPU-port LAG, more than four members, and membership overlap are rejected.

## Hardware validation requested

I do not have an RTL8372N/RTL8373 board available in this environment. Please test on hardware with two or more LAN ports:

- create a bond with `mode=balance-xor` or `mode=802.3ad` and a supported hash policy;
- verify that DSA accepts the ports and `/proc/net/bonding/*` reports the expected slaves;
- generate multiple independent flows and confirm distribution with switch counters or a peer-side capture;
- remove the bond and confirm the member ports return to normal forwarding;
- verify that CPU-port LAG and a fifth member are rejected without changing an existing group.

The software-side mapping and error handling are the reason for proposing this for review; hardware forwarding behaviour still needs confirmation on the target board.

### DIFF
--- a/package/kernel/rtl837x/src/rtl837x_common.h
+++ b/package/kernel/rtl837x/src/rtl837x_common.h
@@ -20,6 +20,7 @@
 #include "./rtk-api/vlan.h"
 #include "./rtk-api/chip.h"
 #include "./rtk-api/eee.h"
+#include "./rtk-api/trunk.h"
 #include "./rtk-api/rma.h"
 #include "./rtk-api/cpuTag.h"
 #include "./rtk-api/mib.h"
@@ -75,6 +76,7 @@ struct rtk_gsw {
 	struct regmap		*map;
 	struct regmap		*map_nolock;
 	struct mutex		map_lock;
+	struct mutex		feature_lock;
 
 	struct gpio_desc *reset_pin;
 	int mdio_addr;
@@ -99,6 +101,8 @@ struct rtk_gsw {
 	unsigned int dsa_num_ports;
 	bool dsa_registered;
 	struct dsa_switch ds;
+	u32 lag_members[TRUNK_GROUP_END];
+	u32 lag_hash_mask[TRUNK_GROUP_END];
 	struct net_device *bridge_dev[RTK_MAX_NUM_OF_PORT];
 	bool port_enabled[RTK_MAX_NUM_OF_PORT];
 	struct net_device *ethernet_master;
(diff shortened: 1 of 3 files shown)


MNeroba:
Adding a concise rationale and validation request for review.

Confidence is approximately 95% for the software integration. The RTL837x SDK already exposes the required trunk-group, member-mask, and distribution-algorithm operations, and the DSA callbacks map directly to those operations. The implementation also enforces the hardware limits, rejects CPU-port and overlapping membership, and keeps the configured state serialized.

This is intentionally static hardware aggregation only. It does not claim LACP, CPU-port aggregation, or FDB distribution beyond what the existing switch SDK provides.

Please validate on a BE9300 with two supported user ports: L2/L23/L34 hash selection, traffic distribution across members, member/link failure behavior, reboot and teardown cleanup, and interaction with the tag_8021q CPU path.

perceival:
Reviewed alongside #47, which is a superset of this and conflicts with it textually —
only one can land, and I'd rather it be #47. The `dsa_lag_id()` one-based/already-unmapped
defect and the `group >= TRUNK_GROUP_END` off-by-one apply to both; details are in my
comment on #47.

Suggest closing this in favour of a fixed #47.

MNeroba:
Closing this duplicate in favor of #47. After review, #47 is the superset and now contains the corrected DSA LAG ID mapping, active-member handling, rollback, and fallback error semantics. Consolidating the work into one PR keeps the review and eventual validation focused.

PR #47, opened 2026-08-31
rtl837x: sync LAG members with DSA link state

MNeroba:
## Summary

This PR supersedes #46 and contains the complete RTL837x DSA LAG
implementation together with `lag_tx_enabled` synchronization.

- Adds the base RTL837x DSA LAG support.
- Synchronizes the hardware LAG transmit state with DSA link state through
  `lag_tx_enabled`.
- Keeps member and hardware state transitions consistent when links go up or
  down.
- Maps one-based DSA LAG IDs to zero-based RTL837x trunk groups; DSA LAG ID 4
  maps to hardware group 3.
- A zero/unassigned DSA LAG ID returns `-EOPNOTSUPP` so DSA can use a software
  LAG; an out-of-range port or LAG ID remains an invalid request.
- Clears hardware trunk member and hash state during setup and teardown.
- #46 was closed in favor of this PR; #47 is self-contained and does not
  depend on a stacked #46 branch.

## Why this is correct

The implementation keeps the hardware LAG transmit decision aligned with the
DSA-reported member link state, including transitions after link changes. The
complete implementation is reviewed as one PR so the base LAG plumbing and
state synchronization are evaluated together. Unsupported software-LAG cases
are reported with `-EOPNOTSUPP`, while malformed hardware port/LAG IDs retain
`-EINVAL`.

The current implementation deliberately does not claim that standalone
`tag_8021q` traffic is safe through a hardware trunk. That datapath question
is called out explicitly below and remains a hardware gate.

## Hardware validation

Keep this PR in Draft until RTL8372N hardware validation is complete:

- Create a LAG with two RTL8372N ports.
- Verify traffic continues when both members are up.
- Bring one member down and verify traffic continues through the remaining
  member.
- Restore the member and verify it rejoins the LAG.
- Exercise repeated link flaps and confirm no stale `lag_tx_enabled` state.
- Verify CPU-injected `tag_8021q` traffic when the hardware trunk redirects a
  frame to a member different from the port encoded by the TX VID; confirm
  VLAN membership does not drop or leak the redirected frame.
- Verify creation and use of DSA LAG ID 4 and confirm it maps to hardware
  trunk group 3, including the one-based to zero-based conversion.

The CPU-injected `tag_8021q` trunk-redirection case is the key datapath test
before using this change on a live LAN.

## Validation

  resolution; no unrelated source changes were added.
- `git diff --check`: passed.
  warnings for the final patch.
- No package build was completed on this macOS host; OpenWrt requires a Linux
  build environment with supported GNU make.

### DIFF
--- a/package/kernel/rtl837x/src/rtl837x_common.h
+++ b/package/kernel/rtl837x/src/rtl837x_common.h
@@ -23,6 +23,7 @@
 #include "./rtk-api/vlan.h"
 #include "./rtk-api/chip.h"
 #include "./rtk-api/eee.h"
+#include "./rtk-api/trunk.h"
 #include "./rtk-api/rma.h"
 #include "./rtk-api/cpuTag.h"
 #include "./rtk-api/mib.h"
@@ -99,6 +100,7 @@ struct rtk_gsw {
 	struct regmap		*map_nolock;
 	struct mutex		map_lock;
 	struct mutex		flood_lock;
+	struct mutex		feature_lock;
 
 	struct gpio_desc *reset_pin;
 	int mdio_addr;
@@ -123,6 +125,8 @@ struct rtk_gsw {
 	unsigned int dsa_num_ports;
 	bool dsa_registered;
 	struct dsa_switch ds;
+	u32 lag_members[TRUNK_GROUP_END];
+	u32 lag_hash_mask[TRUNK_GROUP_END];
 	struct net_device *bridge_dev[RTK_MAX_NUM_OF_PORT];
 	/* Hardware-indexed mask of ports with BR_ISOLATED enabled. */
 	u32 isolated_port_mask;
--- a/package/kernel/rtl837x/src/rtl837x_common.h
+++ b/package/kernel/rtl837x/src/rtl837x_common.h
@@ -126,6 +126,7 @@ struct rtk_gsw {
 	bool dsa_registered;
 	struct dsa_switch ds;
 	u32 lag_members[TRUNK_GROUP_END];
+	u32 lag_active_members[TRUNK_GROUP_END];
 	u32 lag_hash_mask[TRUNK_GROUP_END];
 	struct net_device *bridge_dev[RTK_MAX_NUM_OF_PORT];
 	/* Hardware-indexed mask of ports with BR_ISOLATED enabled. */
--- a/package/kernel/rtl837x/src/rtl837x_dsa_ops.c
+++ b/package/kernel/rtl837x/src/rtl837x_dsa_ops.c
@@ -400,7 +400,7 @@ static int rtl837x_lag_set_members(struct rtk_gsw *gsw, int group,
 	return rtl837x_to_errno(rtk_trunk_port_set(group, &portmask));
 }
 
-static int rtl837x_lag_set_active_members(struct rtk_gsw *gsw, int group)
+static u32 rtl837x_lag_get_active_members(struct rtk_gsw *gsw, int group)
 {
 	u32 active_members = 0;
 	int port;
@@ -416,11 +416,34 @@ static int rtl837x_lag_set_active_members(struct rtk_gsw *gsw, int group)
 			active_members |= BIT(port);
 	}
 
+	return active_members;
+}
+
+static int rtl837x_lag_set_active_members(struct rtk_gsw *gsw, int group)
+{
+	u32 active_members = rtl837x_lag_get_active_members(gsw, group);
+	int ret;
+
 	/* DSA keeps lag_tx_enabled in sync with link state and the LACP
 	 * driver's transmit permission.  The RTL837x trunk API only accepts
-	 * a member mask, so mirror that state into the hardware candidate set.
+	 * one member mask, so use it as the active candidate set while keeping
+	 * lag_members as the configured DSA membership.
 	 */
-	return rtl837x_lag_set_members(gsw, group, active_members);
+	ret = rtl837x_lag_set_members(gsw, group, active_members);
+	if (!ret)
+		gsw->lag_active_members[group] = active_members;
+
+	return ret;
+}
+
+static int rtl837x_lag_group_from_lag(const struct dsa_switch *ds,
+					     const struct dsa_lag *lag)
+{
+	/* DSA LAG IDs are one-based; the RTL837x trunk table is zero-based. */
+	if (!lag->id || lag->id > ds->num_lag_ids)
+		return -EINVAL;
+
+	return lag->id - 1;
 }
 
 static int rtl837x_port_lag_change(struct dsa_switch *ds, int port)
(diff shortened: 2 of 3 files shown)


MNeroba:
Adding the rationale and the remaining validation boundary.

Confidence is approximately 90% for the DSA integration. The driver mirrors the DSA link-state permission exposed as lag_tx_enabled through port_lag_change and programs the active member mask without changing the configured LAG membership. This follows the intended separation where Linux bonding handles 802.3ad negotiation and the switch driver controls which configured members are eligible for transmission.

This remains a draft because neither LACP partner negotiation nor the RTL8372N behavior has been observed on hardware. Please validate an 802.3ad bond with two BE9300 ports, member activation and removal, link-down recovery, traffic distribution, and the no-active-member case.

perceival:
The unwind and locking here are careful, and the `dsa_to_port()` NULL check is right —
but there's a bug that ordinary bond teardown hits every time, and it's shared with #46.
Since #47 is a superset of #46 and the two conflict textually, I'd fix this one and close
#46.

**`dsa_lag_id()` is one-based, and it's already unmapped by the time you look it up.**
`dsa_port_lag_leave()` (net/dsa/port.c:696) calls `dsa_port_lag_destroy(dp)` *before*
firing `DSA_NOTIFIER_LAG_LEAVE`. For the last member the refcount hits zero,
`dsa_lag_unmap()` NULLs `dst->lags[id-1]`, so `dsa_lag_id(ds->dst, lag.dev)` returns
`-ENODEV`. `rtl837x_port_lag_leave()` then bails with `-EINVAL` before ever calling
`rtk_trunk_port_set()`, and DSA only logs it. Net effect: `RTL8373_TRK_MBR_CTRL` still
lists the port and `gsw->lag_members[group]` still has its bit — hardware and shadow
state permanently desynced by a plain `ip link set lanN nomaster`. A later re-join then
either hits the `-EBUSY` scan or silently returns 0 without programming anything.

Fix is small: use the `lag.id - 1` already handed to you in the `struct dsa_lag`
argument and drop the `dsa_lag_id()` lookup entirely.

**Off-by-one on the group space.** `ds->num_lag_ids = TRUNK_GROUP_END` (4) yields ids
1..4, but the bound check `group >= TRUNK_GROUP_END` rejects id 4 — so the 4th bond
always fails and hardware `TRUNK_GROUP0` is never used (`trunk_group_mask == 0x0f`).

**Two writers disagree about the register.** `port_lag_join` writes the full member set,
while `port_lag_change`/`lag_set_active_members` write only the `lag_tx_enabled` subset
to the same register — so a join silently re-admits a link-down member until the next
CHANGELOWERSTATE. And `port_lag_leave` writes the full remaining set then immediately
overwrites it with the active set, while its unwind restores `old_members`, a state the
change path would never produce. Please pick one meaning for that register.

**Error codes.** `net/dsa/user.c:3008` only turns `-EOPNOTSUPP` into a graceful
software-LAG fallback; `-EINVAL`/`-EBUSY`/`-ENOSPC`/`-EIO` abort the enslavement outright
instead of degrading.

**Before this lands anywhere near a live LAN**, it needs a bench test for one thing the
diff can't answer: hardware trunking rewrites egress port selection, but this driver's
forwarding is entirely `tag_8021q` VLAN-based. A CPU-injected frame tagged with lan1's TX
VID that the trunk hash redirects to lan2 is not a member of that VID. I don't know what
the chip does there, and neither of us should guess.

MNeroba:
Thanks for the review. I updated the branch in commit `71b2bae378`.

The follow-up addresses the four points as follows:

- DSA LAG IDs are one-based while the RTL837x trunk table is zero-based, so the callback now uses `lag.id - 1` and accepts hardware groups 0..3 (DSA IDs 1..4). This avoids both the off-by-one access and rejecting the fourth hardware group.
- `lag_members` is now explicitly the configured DSA membership, while `lag_active_members` records the last mask successfully written to hardware. Join, leave, and link-state change all use the same `lag_tx_enabled`-filtered mask; no callback writes the full configured mask behind the active-member path.
- Join/leave now update the shadow state before programming the active mask, matching the established RTL83xx pattern. On a programming failure the software state, active mask, and newly-created hash policy are restored where possible, with an error if restoration itself fails.
- Resource/representation limitations (port already used by another hardware LAG, four-port limit, and incompatible hash policy) return `-EOPNOTSUPP` with an extack so DSA can fall back to software LAG. Actual MDIO/programming failures remain errors and are not silently downgraded.

This follows the DSA contract documented in [`include/net/dsa.h`](https://github.com/torvalds/linux/blob/v6.18/include/net/dsa.h#L106-L194) and the active-candidate handling used by OpenWrt's RTL83xx driver ([`dsa.c`](https://github.com/openwrt/openwrt/blob/main/target/linux/realtek/files-6.18/drivers/net/dsa/rtl83xx/dsa.c#L2390-L2415), [`common.c`](https://github.com/openwrt/openwrt/blob/main/target/linux/realtek/files-6.18/drivers/net/dsa/rtl83xx/common.c#L494-L583)).

Static confidence in the ID mapping, state model, and error handling is about 96%. Hardware validation is still needed for the RTL8372N-specific datapath: two/four-port LAG creation, LACP link up/down transitions, and especially CPU-injected `tag_8021q` frames whose VID/egress selection must remain compatible with the trunk candidate mask.

perceival:
Reviewed against 6.18.39. The groundwork is right: `ds->num_lag_ids = TRUNK_GROUP_END` (4) matches the `RTL8373_TRK_MBR_CTRL_ADDR()` index range, the one-based-to-zero-based conversion is correct, and the hash-mask constants and 4-port limit match the DAL. I also checked the lifetime question — `dsa_port_lag_leave()` copies `info.lag` *before* `dsa_port_lag_destroy()`, so reading `lag.id` in the leave callback is not a use-after-free.

Two blockers.

**1. Wrong errno when no LAG ID is available — this breaks the documented software-LAG fallback.**

`dsa_lag_map()` leaves `lag->id == 0` when the tree's IDs are exhausted, and the kernel says explicitly what a driver must do (`net/dsa/dsa.c`):

> No IDs left, which is OK. Some drivers do not need it. The ones that do, e.g. mv88e6xxx, will discover that dsa_lag_id returns an error for this device when joining the LAG. The driver can then return **-EOPNOTSUPP** back to DSA, which will fall back to a software LAG.

`rtl837x_lag_group_from_lag()` returns `-EINVAL` for `!lag->id`, and `port_lag_join` propagates it. `dsa_user_changeupper()` does not swallow `-EINVAL`, so creating a fifth bond and enslaving a port fails outright instead of quietly becoming a software bond. Same applies to the `!rtl837x_user_port()` rejection.

**2. Hardware trunk hashing defeats tag_8021q's per-port TX VID for router-originated traffic.**

`vsc73xx_xmit()` uses `dsa_tag_8021q_standalone_vid(dp)` for every frame without `offload_fwd_mark` — i.e. all locally-generated *and all routed* traffic. That VID's member set is only `{that port, CPU}`, since `rtl837x_tag_8021q_vlan_add()` does `mbr |= BIT(port)`. Once the trunk is programmed the RTL8373 re-hashes the frame onto whichever member the hash picks; if that is the *other* member, VLAN egress filtering drops it.

Repro: bond lan1+lan2 (balance-xor), no bridge, then ping a host reachable via the bond from the router itself — roughly half the flows black-hole. Bridged traffic is fine because `offload_fwd_mark` uses the shared bridge VID covering both members, which is exactly why this can look like it works in casual testing.

Fix is either to add every trunk member to each member's standalone VID, or to refuse LAG offload while the port is standalone. The PR body notes this is untested on hardware; I think this is the thing to test.

**3. (minor)** Neither `rtl837x_setup()` nor `rtl837x_teardown()` clears `TRK_MBR_CTRL`/`TRK_HASH_CTRL` — only the shadow arrays are memset. A module unload/reload without a GPIO reset leaves the hardware trunking while the driver believes no LAG exists.

**4. (note, not a defect)** `hweight32(...) >= 4` rejecting the 5th port with `-EOPNOTSUPP` yields a bond with 4 hardware members plus one software member. mv88e6xxx has the same shape, so I'm not calling it a bug — but combined with finding 2 it deserves a hardware test.

MNeroba:
Pushed `d3375f6d90` to address the LAG follow-up: an unavailable/zero DSA LAG ID now returns `-EOPNOTSUPP` for software fallback, setup and teardown clear the member/hash hardware state, and the existing ID4-to-group3 mapping is retained. The body keeps the required `tag_8021q` CPU-injected redirection test, so this remains a Draft pending bench validation.

MNeroba:
(`a75f06ba2f`) and pushed `f143ca1095`.

The rebased diff retains the reviewed fixes: zero/unassigned DSA LAG IDs use
`-EOPNOTSUPP` for software-LAG fallback, DSA IDs 1..4 map to hardware groups
0..3, active-member programming is consistent, and setup/teardown clear
hardware trunk state.

The standalone CPU-injected `tag_8021q` redirection case remains an explicit
hardware gate; no source-level assumption was added to paper over it.
`git diff --check` and checkpatch pass, and no hardware/build success is
claimed. The PR therefore remains Draft.

PR #49, opened 2026-08-31
rtl837x: offload egress port rate policing through DSA

MNeroba:
## Summary

Implement an egress-only `FLOW_ACTION_POLICE` offload for RTL8373 through the
DSA `cls_flower` callbacks. A supported rule is one unconditional,
port-wide byte-rate drop policer with an explicitly supplied byte burst.
Ingress police offload is deliberately disabled in this PR.

## RTL8373 mapping

The egress path programs the selected port's rate with the existing
rate-only DAL/RTK operation `rtk_rate_egrBwCtrlRate_set()` and its byte burst
with the new `rtk_rate_egrBwCtrlBurst_set()` mapper operation, then enables
the port meter. The corresponding get operation is also exposed for register
readback.

`FLOW_ACTION_POLICE.police.burst` is passed directly in bytes to
`EGBW_PORT_CTRL.BURST`. The field is 16 bits, so values greater than
`0xffff` are rejected. The setter reads `EGBW_LB_CTRL.TKN` and rejects a
burst smaller than `3 * TKN`; neither condition is silently clamped or
rounded. Because `EGBW_PORT_CTRL` is a logical 64-bit register while the
RTL8373 regmap accesses one 32-bit word at a time, the implementation
accesses the upper word at `EGBW_PORT_CTRL_ADDR(port) + 4` with a 16-bit mask.

The existing rate conversion is retained: Linux bytes/s are converted with
`bytes_per_second / 125` and rounded down to the RTL8373 16-kbit/s rate step,
as accepted in the earlier review. Exact no-clamp validation in this PR
applies to the byte burst field.

The egress rate-only DAL operation writes only the per-port rate field and
does not touch the chip-global `EGBW_CTRL.INC_IFG` setting. The burst setter
also touches only the selected port field.

A port limiter is owned by the installing `cls->cookie`. A different filter
cannot replace an existing port limiter, while a repeated add for the same
cookie updates the rate and burst in place and attempts to restore the
previous configuration if programming the replacement fails. Delete and
teardown clear a reservation only after a successful disable; failed cleanup
is recorded and retried before another owner can use the port.

## Scope and rejected semantics

This PR accepts only one egress, port-wide, byte-rate police action with
`conform-exceed drop` and `notexceed accept`. It rejects ingress police,
selective/non-empty flower matches, non-zero chains, packet-rate/peak-rate/
average-rate policing, overhead, explicit MTU enforcement, unsupported
actions, and missing/zero byte bursts. Statistics are not claimed.

Ingress burst mapping remains a separate follow-up. The generated RTL8373
registers expose `IGBW_PORT_BURST_CTRL.HIGH_ON/HIGH_OFF`, but the semantics
of `IGBW_PORT_DROP_CTRL.DROP_THR` while flow control is disabled are not
resolved well enough to promise Linux police behavior.

## Example

`32k` is used as the review and hardware-test example because it fits the
16-bit egress field without truncation:

```sh
tc qdisc add dev lan2 clsact
tc filter add dev lan2 egress protocol all flower skip_sw \
    action police rate 100mbit burst 32k conform-exceed drop
```

## Validation

  resolution; no ingress burst guess or unrelated source change was added.
- Final diff remains limited to the RTL8373 rate mapper/API and DSA policing
  implementation; no ingress police callback is registered.
- `git diff --check`: passed.
  warnings for the final patch.
- The accepted `/125` conversion and 16-kbit/s rate quantization are
  unchanged; exact no-clamp validation applies to burst only.
- No package build was completed during this refresh on the macOS host; no
  successful package compilation is claimed.

Keep this PR in Draft pending the existing maintainer and hardware review
gates. Hardware validation should verify egress rule installation with
`burst 32k`, exact BURST register readback, rejection below `3*TKN` and above
`0xffff` without silent clamp, accepted 16-kbit/s rate quantization,
unchanged chip-global `EGBW_CTRL.INC_IFG`, sustained over-limit egress
traffic, cleanup, and rejection of ingress rules.

References: Linux `FLOW_ACTION_POLICE` ([v6.18 `flow_offload.h`](https://raw.githubusercontent.com/torvalds/linux/v6.18/include/net/flow_offload.h)), the RTL8373 generated register definitions ([`rtl8373_reg_definition.h`](https://github.com/MNeroba/openwrt-flint3/blob/feature/rtl837x-rate/package/kernel/rtl837x/src/rtk-api/dal/rtl8373/rtl8373_reg_definition.h#L11706-L11718)), and the chip-specific burst handling in the [upstream RTL930x DSA implementation](https://raw.githubusercontent.com/openwrt/openwrt/main/target/linux/realtek/files-6.18/drivers/net/dsa/rtl83xx/rtl930x.c).


MNeroba:
Adding the rationale and the exact scope of this offload.

Confidence is approximately 95% for the software and SDK mapping. The RTL8373 DAL exposes both port rate registers and their enable bits; the patch adds the missing egress wrapper and uses the existing ingress/egress rate controls. The implementation intentionally accepts only one port-wide byte-rate police action, rejects packet-rate and unsupported semantics, and rejects flower matches so a selective rule cannot be silently widened into a whole-port limiter.

This is rate policing only; it does not claim QoS scheduling or storm-control offload. Please validate ingress and egress enforcement with tc, the effective rate and burst behavior, replace and remove operations, both directions together, link flap/reboot cleanup, and the failure path when the requested rate is outside the RTL837x range.

MNeroba:
## Follow-up research: PPE/NAT and full QoS/storm-control

@perceival, I audited the current OpenWrt, Linux DSA and RTL8373 sources to determine what can be implemented safely without inventing a hardware contract. The result is useful evidence, but not yet a basis for a functional NAT or full QoS/storm-control patch.

### Confirmed in the current tree

- The RTL8373 SDK exposes four ingress storm groups: unknown unicast, unknown multicast, multicast and broadcast. Each has a per-port enable bit and a six-bit shared-meter selector. The register definitions cover ports 0-9 and meters 0-63.
- The QoS API exposes eight priorities, eight queues, per-port priority-to-queue mapping, and per-queue strict/WFQ scheduling.
- PR #50 fixes a prerequisite readback bug: `dal_rtl8373_rate_shareMeter_get()` read IFG into `pIfg_include` and then overwrote it from `regData` before `regData` was read for the meter type.
- The IPQ5332 PPE patch in this tree currently adds SoC-specific ICC, port and EDMA ring topology. It does not add an IPQ5332 flow-table or NAT programming path.

### What is still needed

For PPE/NAT, could you add or point us to an authoritative IPQ5332 source for the PPE flow tables: register layout, key/result format, add/remove/flush lifecycle, counters, miss/eviction handling, port/VSI identity and NAT rewrite fields? A read-only debugfs/register dump from a BE9300 would also be useful if the layout is already known. The upstream Qualcomm PPE series describes routing capability, but presents a base PPE driver; the OpenWrt IPQ807x work demonstrates that a real host-driven NAT implementation needs a separate flow-table layer, DSA `TC_SETUP_FT` plumbing and register readback.

For RTL8373 QoS/storm-control, could you add or point us to the documented meter units, burst semantics and allocation rules? The 64 shared meters are also referenced by ACL policing, so a future implementation needs a common allocator and rollback/readback policy. We also need a decision on the Linux interface: generic DSA provides ingress `port_policer_add/del` and `port_setup_tc`, but no storm-specific callback; the current DSA user netdev defaults to one TX queue, while the 802.1Q tagger derives PCP from queue mapping.

Until these materials or hardware readback are available, I will keep PPE/NAT and full QoS/storm-control out of functional PRs. PR #49 remains limited to port-wide byte-rate policing, and PR #50 remains the isolated shared-meter readback fix.

References:

- Qualcomm PPE overview: https://docs.kernel.org/7.1/networking/device_drivers/ethernet/qualcomm/ppe/ppe.html
- Qualcomm IPQ5332/IPQ9574 PPE series: https://patchew.org/linux/20240110114033.32575-1-quic._5Fluoj%40quicinc.com/
- OpenWrt host-driven PPE/NAT reference: https://github.com/openwrt/openwrt/pull/24806
- Linux DSA TC callbacks: https://raw.githubusercontent.com/torvalds/linux/v6.18/include/net/dsa.h
- Linux DSA TC forwarding path: https://raw.githubusercontent.com/torvalds/linux/v6.18/net/dsa/user.c
- RTL8373 storm API in this fork: https://github.com/MNeroba/openwrt-flint3/blob/fix/rtl8373-sharemeter-get/package/kernel/rtl837x/src/rtk-api/storm.h#L21-L28
- RTL8373 QoS API in this fork: https://github.com/MNeroba/openwrt-flint3/blob/fix/rtl8373-sharemeter-get/package/kernel/rtl837x/src/rtk-api/qos.h#L56-L61

perceival:
The unit arithmetic here is correct — `rate_bytes_ps / 125` is exactly kbit/s, and
`&= ~0xf` matches the DAL's `regData = rate >> 4` 16-kbit/s granularity and
`INBW_CTRL_RATE_MAX`. The new `rtk_rate_egrBwCtrlPortEn_{set,get}` wrappers bind to DAL
functions that already exist, so it builds. But as written the offload can never engage.

**The dissector guard rejects every possible flower rule:**

```c
if (cls->rule->match.dissector && cls->rule->match.dissector->used_keys)
        return -EOPNOTSUPP;
```

`fl_init_dissector()` (net/sched/cls_flower.c:2173-2174) sets
`FLOW_DISSECTOR_KEY_CONTROL` and `FLOW_DISSECTOR_KEY_BASIC` **unconditionally** for every
filter, so `used_keys` is never zero and `rtl837x_cls_flower_add()` returns
`-EOPNOTSUPP` 100% of the time. The intent — refuse to turn a selective match into a
port-wide limiter — is exactly right; the test needs to mask off CONTROL/BASIC and then
check the basic key's masks are actually empty.

**Second issue, latent until the gate is fixed:** `rtl837x_rate_enable()` passes
`ipg = DISABLED` to `rtk_rate_egrBandwidthCtrlRate_set()`, which lands on
`RTL8373_EGBW_CTRL_INC_IFG_OFFSET` — a **chip-global** register
(dal_rtl8373_rate.c:424), not per-port. So adding an egress policer to one port silently
changes IFG accounting for every port's egress shaper.

Also `.cls_flower_stats` is absent, so `tc -s filter` will always report zero.

Happy to take this once the guard actually admits a bare matchall/police rule and the IFG
bit is either left alone or made explicit.

MNeroba:
## Follow-up: egress byte-burst mapping is now implemented

@perceival, the deeper RTL8373 audit resolved the egress side well enough to
complete this PR's supported scope.

### Confirmed mapping

* `EGBW_LB_CTRL.TKN` is read from the shared egress leaky-bucket control
  register before every burst write, and `burst < 3 * TKN` is rejected.
* `EGBW_PORT_CTRL.BURST` is a 16-bit byte field. Linux
  `FLOW_ACTION_POLICE.police.burst` is passed directly; values above `0xffff`
  are rejected rather than clamped.
* The generated register definition describes BURST at logical bit 32 of a
  64-bit `EGBW_PORT_CTRL`, but this driver uses a 32-bit regmap. The DAL
  therefore accesses `RTL8373_EGBW_PORT_CTRL_ADDR(port) + 4` with mask
  `0xffff`, which is the physical upper word and avoids passing the unusable
  `(0xffff << 32)` mask to the 32-bit accessor.
* New `rtk_rate_egrBwCtrlBurst_{set,get}()` wrappers and mapper slots expose
  this operation through the normal DAL/RTK path.
* Egress rate programming continues to use the rate-only DAL operation, so
  `EGBW_CTRL.INC_IFG` remains unchanged. The rate conversion also rejects
  non-exact values instead of rounding down to a 16-kbit/s step.

### Scope decision

The DSA implementation is now explicitly egress-only: ingress police add is
rejected and the ingress limiter state/teardown path was removed. Ingress
burst mapping stays out of this PR because `IGBW_PORT_DROP_CTRL.DROP_THR` in
FC-disabled drop mode still has no sufficiently verified Linux-police
semantic mapping. That is a separate follow-up, not a silent rate-only
offload.

The accepted action is one unconditional port-wide byte-rate policer with a
non-zero byte burst and `conform-exceed drop`/`notexceed accept`. Selective or
non-empty flower matches, unsupported police fields (including explicit MTU),
and unsupported actions remain rejected.

The review example is now:

```sh
tc qdisc add dev lan2 clsact
tc filter add dev lan2 egress protocol all flower skip_sw \
    action police rate 100mbit burst 32k conform-exceed drop
```

The implementation is in commit [`ca955e40ad`](https://github.com/MNeroba/openwrt-flint3/commit/ca955e40ad). `git diff --check` and checkpatch pass. The package build was attempted with both system GNU make and Homebrew `gmake`; this macOS checkout stops at OpenWrt host prerequisites (case-sensitive Linux filesystem and GNU host tools), so live hardware and Linux package-build validation remain to be done. The PR remains Draft until the register readback, IFG-preservation, `3*TKN` boundary, over-limit traffic, and ingress-rejection tests are run on hardware.

MNeroba:
## Follow-up: egress byte-burst mapping is now implemented

@perceival, the deeper RTL8373 audit resolved the egress side well enough to
complete this PR's supported scope.

### Confirmed mapping

* `EGBW_LB_CTRL.TKN` is read from the shared egress leaky-bucket control
  register before every burst write, and `burst < 3 * TKN` is rejected.
* `EGBW_PORT_CTRL.BURST` is a 16-bit byte field. Linux
  `FLOW_ACTION_POLICE.police.burst` is passed directly; values above `0xffff`
  are rejected rather than clamped.
* The generated register definition describes BURST at logical bit 32 of a
  64-bit `EGBW_PORT_CTRL`, but this driver uses a 32-bit regmap. The DAL
  therefore accesses `RTL8373_EGBW_PORT_CTRL_ADDR(port) + 4` with mask
  `0xffff`, which is the physical upper word and avoids passing the unusable
  `(0xffff << 32)` mask to the 32-bit accessor.
* New `rtk_rate_egrBwCtrlBurst_{set,get}()` wrappers and mapper slots expose
  this operation through the normal DAL/RTK path.
* Egress rate programming continues to use the rate-only DAL operation, so
  `EGBW_CTRL.INC_IFG` remains unchanged. The previously reviewed
  `rate_bytes_ps / 125` conversion and round-down to the existing 16-kbit/s
  rate step are retained; exact no-clamp validation here applies to BURST.
  The burst setter also touches only the selected port field.

### Scope decision

The DSA implementation is now explicitly egress-only: ingress police add is
rejected and the ingress limiter state/teardown path was removed. Ingress
burst mapping stays out of this PR because `IGBW_PORT_DROP_CTRL.DROP_THR` in
FC-disabled drop mode still has no sufficiently verified Linux-police
semantic mapping. That is a separate follow-up, not a silent rate-only
offload.

The accepted action is one unconditional port-wide byte-rate policer with a
non-zero byte burst and `conform-exceed drop`/`notexceed accept`. Selective or
non-empty flower matches, unsupported police fields (including explicit MTU),
and unsupported actions remain rejected.

The review example is now:

```sh
tc qdisc add dev lan2 clsact
tc filter add dev lan2 egress protocol all flower skip_sw \
    action police rate 100mbit burst 32k conform-exceed drop
```

The implementation is in commits [`ca955e40ad`](https://github.com/MNeroba/openwrt-flint3/commit/ca955e40ad) and [`139905937d`](https://github.com/MNeroba/openwrt-flint3/commit/139905937d); the latter explicitly retains the previously reviewed rate quantization policy. `git diff --check` and checkpatch pass. The package build was attempted with both system GNU make and Homebrew `gmake`; this macOS checkout stops at OpenWrt host prerequisites (case-sensitive Linux filesystem and GNU host tools), so live hardware and Linux package-build validation remain to be done. The PR remains Draft until the register readback, IFG-preservation, `3*TKN` boundary, over-limit traffic, and ingress-rejection tests are run on hardware.

MNeroba:
## Final audit: egress-only burst offload

@perceival, I rechecked the complete revision after the latest rollback fix
(`95bb009439`) against the previous #49 head (`08f49b98fc`).

1. The delta contains only the eight expected API/DAL/DSA files. There are no
   ingress burst guesses, raw-register writes in the DSA callback, or unrelated
   cleanup changes.
2. Egress `police.burst` is passed directly to `EGBW_PORT_CTRL.BURST`; the DAL
   reads `EGBW_LB_CTRL.TKN` and rejects `burst < 3*TKN` and values above the
   16-bit field without clamping. The selected port's upper register word is
   used for the 64-bit logical register.
3. The rate-only egress DAL operation leaves global `EGBW_CTRL.INC_IFG`
   unchanged. The existing `/125` conversion and round-down to the 16-kbit/s
   rate step are intentionally retained; exact no-clamp validation applies to
   burst, not rate.
4. Ingress police add returns `-EOPNOTSUPP` before programming hardware and no
   ingress limiter reservation/state remains in this PR.
5. Rate, burst, and enable failures all attempt to disable the port meter;
   add releases the reservation on failure, while delete/teardown clear it only
   after successful disable.
6. Implicit `CONTROL`/`BASIC` flower keys are accepted, while selective keys or
   non-empty BASIC masks are rejected, so a selective rule cannot become a
   port-wide limiter.
7. The PR title/body and `burst 32k` example match the egress-only scope. The
   complete delta passes `git diff --check` and checkpatch. The package build
   and hardware readback/traffic tests still require the Linux build host and
   RTL8373 bench hardware noted in the description; current GitHub checks are
   pending.

perceival:
Reviewed against 6.18.39. I went in suspicious of the register arithmetic and it holds up: `RTL8373_REG_ADDR_STEP` is 4 with a byte-addressed map (`EGBW_PORT_CTRL` 0x1C34, next 0x1C3C, 8-byte stride), so `+4` for the BURST word at bits 47:32 is right, and the local `0xFFFF` mask correctly dodges the `0xFFFF << 32` UB in the generated header. `setAsicRegBits`/`getAsicRegBits` shift by `__ffs(mask)`, RATE and EN share a word but `regmap_update_bits` preserves EN, and the `used_keys` stripping of CONTROL+BASIC correctly identifies a bare port-wide flower rule. Callback signatures match `include/net/dsa.h`.

The problem is that the reservation is per-port but not tied to the filter that owns it.

**1. An unrelated filter's DESTROY tears down the offload.**

`tc_setup_cb_destroy()` calls `__tc_setup_cb_call()` unconditionally — there is no "was this filter actually in hardware" gate (verified in `net/sched/cls_api.c`), and `fl_hw_destroy_filter()` runs for any non-`skip_hw` filter.

```
tc filter add dev lan2 egress prio 1 flower action police rate 100mbit burst 32k   # offloaded, meter on
tc filter add dev lan2 egress prio 2 flower ip_proto tcp action drop               # driver returns -EOPNOTSUPP,
                                                                                   # swallowed, installs software-only
tc filter del dev lan2 egress prio 2                                               # <-- disables the port meter
```

The prio-1 rate limit is gone while `tc filter show` still lists it as offloaded.

**2. Replacing an existing rule silently removes the rate limit.**

`fl_change()` calls `fl_hw_replace_filter(fnew)` before `fl_hw_destroy_filter(fold)`. With the port bit already set the add returns `-EOPNOTSUPP` (swallowed without `skip_sw`), then the old filter's destroy disables the meter. Re-issuing the PR's own example with `rate 200mbit` at the same prio leaves no hardware policing at all and reports no error. With `skip_sw` it instead fails outright, so there is no way to change a rate in place.

Both are fixed by storing the owning `cls->cookie` per port, matching it in `cls_flower_del`, and permitting a replace when the cookie matches.

**3. (minor)** If `rtl837x_rate_disable()` fails in `cls_flower_del`, the bit is deliberately left set "so cleanup can be retried" — but tc has already removed the filter, so no retry path exists. The port stays rate-limited forever and every later add on it returns `-EOPNOTSUPP`.

**4. (minor)** `rtl837x_setup()` zeroes `rate_egress_mask` without clearing `EGBW_PORT_CTRL.EN` in hardware — same stale-state class as #47.

MNeroba:
Pushed `115fa89f8d` to make failed egress-limiter cleanup retryable without disabling a healthy limiter owned by another filter. A failed delete records pending cleanup; a later add retries the disable and only then releases the old ownership. Initial programming failures also record a failed cleanup attempt when the hardware disable itself fails. The egress-only scope, exact burst checks, preserved IFG setting, and existing 16-kbit/s rate quantization remain unchanged.

MNeroba:
(`a75f06ba2f`) and pushed `dea7744641`.

The rebased diff retains the reviewed egress-only scope: exact byte-burst
bounds and `3*TKN` validation, preserved chip-global IFG, owner-cookie
protection for replace/delete, retryable cleanup, and setup-time disabling of
stale egress meters. The accepted `/125` conversion and 16-kbit/s rate
quantization are unchanged; exact representation checks apply to burst only.

`git diff --check` and checkpatch pass. Hardware register/readback and traffic
validation remain outstanding, so this PR remains Draft.

PR #52, opened 2026-09-01
rtl837x: clean probe state on initialization failure

MNeroba:
## Summary

Clean up the two RTL837x probe error paths that run after driver state has been installed:

- remove the manual `devm_kfree()` calls;
- clear `dev_set_drvdata()` before returning from probe failure.

The existing `rtl837x_clear_global_priv()` calls are retained.

## Why this is correct

`gsw` is allocated with devres, so probe failure already releases it during
normal devres unwinding. Calling `devm_kfree()` directly in the error path is
unnecessary. It is not an active use-after-free in the current code because
the regmap release path does not dereference `lock_arg`, but retaining a
manual free makes the lifetime assumptions fragile for future devm-managed
contexts created for the object. Removing it preserves the normal resource
lifetime.

The probe stores `gsw` in both the module-global private pointer and device
driver data before hardware initialization and DSA registration. The global
pointer is already cleared on these failures; clearing `drvdata` as well
prevents the failed device from retaining a pointer to probe state that will be
released by devres. This is defense in depth for the failed-probe path.

This is intentionally limited to probe cleanup. It does not change the error
policy, DSA registration order, hardware initialization, or normal remove path.

## Validation

- `git diff --check`: passed.
- Reviewed both failure paths against the devres allocation and the existing `rtl837x_clear_global_priv()` helper.
- The branch name intentionally has no `codex/` prefix.

No router was required to establish the lifetime and pointer-cleanup
correctness. Please review the devres ordering and confirm that both
failed-probe paths should use the same cleanup policy.

### DIFF
--- a/package/kernel/rtl837x/src/rtl837x_mdio.c
+++ b/package/kernel/rtl837x/src/rtl837x_mdio.c
@@ -938,9 +938,9 @@ static int rtl837x_dsa_probe(struct mdio_device *mdiodev)
 	{
 		dev_err(gsw->dev, "rtl8372n_hw_init failed, ret=%d\n",ret);
 		rtl837x_clear_global_priv(gsw);
+		dev_set_drvdata(dev, NULL);
 		if (master)
 			dev_put(master);
-		devm_kfree(dev, gsw);
 		return -ENODEV;
 	}
 
@@ -951,9 +951,9 @@ static int rtl837x_dsa_probe(struct mdio_device *mdiodev)
 	if (ret){
 		dev_err(gsw->dev, "rtl837x_dsa_register failed, ret=%d\n", ret);
 		rtl837x_clear_global_priv(gsw);
+		dev_set_drvdata(dev, NULL);
 		if (master)
 			dev_put(master);
-		devm_kfree(dev, gsw);
 		return ret;
 	}
 


MNeroba:
### Author note

This follows the probe-lifetime concerns raised in #18 and #19 and is deliberately limited to cleanup after a failure has already occurred.

`gsw` is allocated through devres, so manual `devm_kfree()` in these error paths is unnecessary and can release the object before devm-managed users of it have unwound. The patch therefore leaves ownership to devres. It also clears `drvdata` alongside the existing `rtl837x_clear_global_priv()` call, so neither driver-visible pointer retains failed probe state after the device is rejected.

No normal initialization, DSA registration order, error policy, or remove path is changed. The source-level lifetime correction is at least 95% confident; please review it against the current probe registration/unwind ordering and confirm that no failed-probe caller needs `drvdata` to remain set.

perceival:
**Merging** — both changes are right, though I'd like to correct the stated rationale, because I initially believed the stronger version of it myself and it does not survive checking.

**Removing `devm_kfree(dev, gsw)`** is correct on redundancy grounds: `gsw` is `devm_kzalloc`'d, so probe failure releases it during normal devres unwinding and the manual free is pointless.

But the implied use-after-free does **not** currently happen. It is true that both `devm_regmap_init()` calls pass `gsw` as `lock_arg` and that `gsw` is registered in devres *before* the regmaps, so `devm_kfree` did free it ahead of `devm_regmap_release()`. However `regmap_exit()` never touches `lock_arg` — it only `mutex_destroy()`s its own mutex, and only when `map->lock == regmap_lock_mutex`, which is not the case here since this driver installs custom lock ops. So the hazard is **latent, not live**: it would become a real UAF the moment anything touched the regmap or `gsw->map_lock` during unwind. Worth fixing for exactly that reason, but the commit message shouldn't claim it fixes an active use-after-free.

**Adding `dev_set_drvdata(dev, NULL)`** is worth having as defence in depth, with the same caveat. `device_unbind_cleanup()` in the driver core already does `dev_set_drvdata(dev, NULL)` on probe failure, and `dev->driver` is cleared so neither `rtl837x_dsa_remove()` nor `rtl837x_mdio_shutdown()` can fire. So this is not a live crash fix on its own — but combined with the old `devm_kfree()` it *was* a genuine dangling pointer, and both those callbacks start with `dev_get_drvdata()` + `if (!gsw) return`, which is exactly the shape that bites.

Cleanup ordering on both paths is otherwise fine, and nothing is leaked at either failure point.

Two pre-existing issues, not introduced here and not blocking: the `hw_init` path returns a hardcoded `-ENODEV` instead of `ret`, and `gsw->ethernet_master` is left set after `dev_put(master)`.

PR #53, opened 2026-09-01
rtl837x: avoid overriding bridge STP on port enable

MNeroba:
## Summary

Remove the RTL837x DSA driver callbacks that force ports to `BR_STATE_FORWARDING` during port enable and to `BR_STATE_DISABLED` during port disable.

The patch also removes the unused `port_enabled[]` state from `struct rtk_gsw` and its setup bookkeeping.

## Why this is correct

DSA already owns the port lifecycle and applies the generic STP state for standalone ports. For bridged ports, the bridge STP state is the authoritative state and must not be overwritten by a driver-level enable callback. The current callbacks therefore race with or override bridge-controlled STP transitions.

The driver still exposes `port_stp_state_set`, which is the appropriate DSA hook for programming hardware STP state. No hardware initialization, VLAN handling, link handling, or bridge bookkeeping is changed.

## Validation

- `git diff --check`: passed.
- Confirmed `port_enabled[]` had no readers and was only written by the removed callbacks and setup code.
- Reviewed the Linux 6.18 DSA port lifecycle: generic forwarding and disabled states are only applied when the port is not attached to a bridge.

No router was required for this control-flow and dead-state cleanup. Please review the STP ownership boundary and confirm that the existing `port_stp_state_set` path is sufficient for the supported topology.

## Package build validation

### DIFF
--- a/package/kernel/rtl837x/src/rtl837x_common.h
+++ b/package/kernel/rtl837x/src/rtl837x_common.h
@@ -100,7 +100,6 @@ struct rtk_gsw {
 	bool dsa_registered;
 	struct dsa_switch ds;
 	struct net_device *bridge_dev[RTK_MAX_NUM_OF_PORT];
-	bool port_enabled[RTK_MAX_NUM_OF_PORT];
 	struct net_device *ethernet_master;
 	struct sfp_bus *sfp_bus;
 
--- a/package/kernel/rtl837x/src/rtl837x_dsa_ops.c
+++ b/package/kernel/rtl837x/src/rtl837x_dsa_ops.c
@@ -394,12 +394,10 @@ static int rtl837x_setup(struct dsa_switch *ds)
 		return ret;
 
 	memset(gsw->bridge_dev, 0, sizeof(gsw->bridge_dev));
-	memset(gsw->port_enabled, 0, sizeof(gsw->port_enabled));
 	memset(gsw->tag8021q_pvid, 0, sizeof(gsw->tag8021q_pvid));
 	memset(gsw->tag8021q_pvid_valid, 0, sizeof(gsw->tag8021q_pvid_valid));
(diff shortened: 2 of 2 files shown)


perceival:
**Merging — and it removes a real bug, not just redundant code.** Verified in `net/dsa/port.c` (6.18.39): the core deliberately skips the generic state transition for bridged ports, in both directions:

```c
dsa_port_enable_rt():   if (ds->ops->port_enable) ...;  if (!dp->bridge) dsa_port_set_state_now(dp, BR_STATE_FORWARDING, false);
dsa_port_disable_rt():  if (!dp->bridge) dsa_port_set_state_now(dp, BR_STATE_DISABLED, false);  if (ds->ops->port_disable) ...;
```

The driver's hooks were doing it unconditionally, so on a bridged port they overrode the bridge behind its back:

```
ip link add br0 type bridge stp_state 1
ip link set lan1 master br0          # bridge -> BLOCKING, driver programs MSTP_BLOCKING
ip link set lan1 down; ip link set lan1 up
                                     # port_enable forces MSTP_FORWARDING
```

Hardware then forwards while both the bridge and `dp->stp_state` still say BLOCKING — a transient L2 loop — and because `dp->stp_state` never saw the change, DSA's fast-age heuristic makes the wrong call on the next real transition. `port_disable` had the mirror-image problem, forcing `MSTP_DISABLE` on a port the bridge believed was FORWARDING.

Removal is safe in the remaining cases: CPU/DSA ports and standalone user ports have no bridge, so the core sets FORWARDING/DISABLED itself; bridged ports get their state via switchdev (`br_stp_enable_port()` → `br_port_state_selection()`, immediate FORWARDING under `BR_NO_STP`, which is the OpenWrt default). Dropping `port_enabled` is fine too — nothing reads it anywhere, `rtl837x_swconfig.c` included.

PR #54, opened 2026-09-01
rtl837x: keep VLAN filtering and PVID transitions consistent

MNeroba:
## Summary

Make VLAN filtering transitions use the requested mode when selecting the PVID source, and restore the previous ingress-filtering state if the PVID update fails.

The existing PVID helper remains the compatibility wrapper for VLAN add and delete paths, while the transition path now passes the target mode explicitly.

## Why this is correct

DSA invokes the switch driver callback with the new `vlan_filtering` value before it updates the DSA port state. Reading `dsa_port_is_vlan_filtering()` inside that callback therefore describes the old mode. The old-mode read can select the tag_8021q PVID while enabling bridge VLAN filtering, or the bridge PVID while disabling it.

The callback now commits the PVID using the requested mode. If that commit fails after ingress filtering has already changed, the driver attempts to restore the old ingress-filtering setting and returns the original PVID error. A failed rollback is reported explicitly. VLAN shadow state is updated only after the PVID write succeeds.

This is intentionally limited to the transition ordering and rollback. It does not change VLAN table membership, tag_8021q VID allocation, or the normal VLAN add and delete paths.

## Validation

- `git diff --check`: passed.
- Reviewed the Linux 6.18 DSA callback ordering and confirmed that the callback receives the requested mode before `dp->vlan_filtering` is updated.
- Confirmed existing VLAN add and delete callers retain the current-state wrapper.

No router was required for this state-ordering and rollback fix. Please review the failure sequence and confirm that restoring ingress filtering is the desired hardware rollback when the PVID write fails.

## Package build validation

### DIFF
--- a/package/kernel/rtl837x/src/rtl837x_dsa_ops.c
+++ b/package/kernel/rtl837x/src/rtl837x_dsa_ops.c
@@ -129,14 +129,14 @@ static int rtl837x_open_isolation(struct rtk_gsw *gsw)
 	return 0;
 }
 
-static int rtl837x_commit_pvid(struct rtk_gsw *gsw, int port)
+static int rtl837x_commit_pvid_for_mode(struct rtk_gsw *gsw, int port,
+					bool vlan_filtering)
 {
-	struct dsa_port *dp = dsa_to_port(&gsw->ds, port);
 	bool valid = gsw->tag8021q_pvid_valid[port];
 	u16 vid = gsw->tag8021q_pvid[port];
 	int ret;
 
-	if (dsa_port_is_vlan_filtering(dp)) {
+	if (vlan_filtering) {
 		vid = gsw->bridge_pvid[port];
 		valid = gsw->bridge_pvid_valid[port];
 	}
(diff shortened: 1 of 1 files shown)


perceival:
**Merging — and this is a real bug fix, not just a refactor.** I verified the premise in the DSA core rather than taking it on trust.

`dsa_port_vlan_filtering()` (`net/dsa/port.c`) calls the driver callback *first* and only updates the port's filtering state afterwards:

```c
err = ds->ops->port_vlan_filtering(ds, dp->index, vlan_filtering, extack);
if (err)
        return err;
/* ... state updated only after this point ... */
```

So inside `rtl837x_port_vlan_filtering()`, `dsa_port_is_vlan_filtering(dp)` still returns the **old** value. The previous code therefore committed the PVID for the mode we were leaving, on every transition: enabling VLAN filtering left the port's PVID as the tag_8021q VID instead of the bridge VID, and disabling it did the reverse — until some unrelated event happened to re-commit. Passing the intended value explicitly is the right fix.

That matters more than it looks on this driver, because the port PVID is what `tag_8021q` uses to identify the source port on ingress. A PVID left in the wrong mode directly affects which netdev DSA attributes received frames to.

The `rtl837x_commit_pvid()` wrapper keeps every existing caller correct, and rolling back `rtk_vlan_portIgrFilterEnable_set()` when the PVID commit fails is a good addition — without it a failure would leave ingress filtering and the PVID describing different modes.

No objections.

PR #55, opened 2026-09-01
rtl837x: propagate debugfs register access errors

MNeroba:
## Summary

Return an error to userspace when RTL837x debugfs register, PHY-MMD, or SerDes
access fails instead of reporting a successful write or formatting an invalid
read value.

The patch also validates the SerDes ID and PHY port on both read and write
paths, keeping the existing input-error behavior for malformed requests.

## Why this is correct

The debugfs handlers call the vendor SDK directly, but the previous
implementation ignored return values from all register access operations. A
failed MDIO transaction could therefore appear successful to the caller, and a
failed read could expose an uninitialized or stale value in the readback
buffer.

Each hardware operation now checks its return value, frees the userspace
buffer, and returns `-EIO` on an SDK failure. Invalid SerDes and PHY indices
are rejected symmetrically before accessing the SDK. Readback buffers are
cleared before each new command, so a malformed command, failed read, or
failed userspace copy cannot expose the previous successful read result. The
`port > 9` check is the explicit RTL837x debug-interface capability limit, not
a substitute for a buffer-safety check. The existing SerDes page dump path
already propagates its access failures and is left unchanged.

This is intentionally limited to debugfs error reporting and bounds
validation. It does not change the vendor register format, command syntax, or
normal DSA paths.

## Validation

- `git diff --check`: passed.
- Confirmed all six direct register access calls in the three write handlers now have checked return values.
- Confirmed readback formatting occurs only after a successful hardware read.
- Confirmed failed commands clear the corresponding readback buffer.
- Confirmed the existing page-dump error path remains intact.

No router was required for this userspace error-path fix. Please review the
errno choice and the symmetric index validation.

## Package build validation

### DIFF
--- a/package/kernel/rtl837x/src/rtl837x_dbg.c
+++ b/package/kernel/rtl837x/src/rtl837x_dbg.c
@@ -9,6 +9,7 @@
 #include <linux/proc_fs.h>
 #include <linux/u64_stats_sync.h>
 #include <linux/slab.h>
+#include <linux/errno.h>
 
 #include "./rtl837x_common.h"
 
@@ -43,6 +44,7 @@ ssize_t _sdsreg_rw_write(struct file *filep, const char __user *ubuf,
 {
 	char *buf;
 	uint32_t sds_id, page, reg, val;
+	int ret;
 
 	buf = memdup_user_nul(ubuf, count);
 	if (IS_ERR(buf))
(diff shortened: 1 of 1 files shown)


perceival:
**Merging with two notes.** The propagation is correct: `RT_ERR_OK == 0` and every other `RT_ERR_*` is nonzero, so `if (ret)` catches both the `-1` `RT_ERR_FAILED` and the positive codes, all six call sites are covered, and the read buffers are only formatted after success. No context problem — these are process-context writes and the RTK lock is a mutex.

**1. The stale-readback hole the PR is chartered to close is still open.** `_buf_rd_reg` and friends are static and are not invalidated on the error path:

```
echo "r 0x1000" > .../reg     # succeeds
echo "r 0x7b20" > .../reg     # fails, -EIO
cat .../reg                   # still prints reg: 0x00001000, val: ...
```

Any script doing `echo … > reg; cat reg` without checking the write result reads the *previous* register's value believing it is the new one — which is precisely the confusion the change is meant to prevent. Clearing or overwriting the buffer before returning `-EIO` would close it.

**2. The new `port > 9` bound on the read path isn't the symmetry it looks like.** `rtk_port_phyReg_set()` takes a *phy mask* (hence the existing `1 << port`), while `rtk_port_phyReg_get()` takes a *phy id* that the SDK only places in a 16-bit register field — there's no array indexing to protect. So this is a capability restriction on a debug interface rather than a safety fix: `echo "r 10 0x1e 0x1234" > phyreg_mmd` now returns `-EFAULT` where it previously performed the MMD read. Harmless on this board (phy ids in use are 0/1), just worth not describing as a bounds fix.

MNeroba:
Pushed `ca327bf121` to invalidate each static debugfs readback buffer before processing a new command. Failed, malformed, or failed-read requests therefore cannot expose the previous successful value. The body also clarifies that the `port > 9` check is an RTL8373 capability restriction, not the memory-safety fix.

PR #56, opened 2026-09-01
rtl837x: return Linux errno from MDIO regmap callbacks

MNeroba:
## Summary

Translate the RTL837x MDIO busy timeout into the Linux `-ETIMEDOUT` errno in
all regmap read and write callback paths.

The existing negative errors returned by the underlying MDIO bus are preserved
unchanged.

## Why this is correct

The regmap callbacks return Linux-style negative errno values to their callers.
`RT_ERR_BUSYWAIT_TIMEOUT` is a positive vendor SDK status code, so returning it
directly violates the callback contract; current callers test nonzero results,
but a positive value can be misinterpreted by other Linux consumers as a
successful or otherwise invalid result.

The timeout condition is now reported as `-ETIMEDOUT` for both the
pre-operation and post-operation busy checks in read and write callbacks.
Bus-level failures continue to pass through, preserving the more specific
error reported by the MDIO bus.

This is intentionally limited to the regmap boundary. It does not alter the
MDIO transaction sequence, locking, timeout detection, or vendor SDK status
handling elsewhere in the driver.

## Validation

- `git diff --check`: passed.
- Confirmed all four busy-timeout assignments in `rtl837x_mdio_read` and `rtl837x_mdio_write` now return `-ETIMEDOUT`.
- Confirmed negative MDIO bus errors remain unchanged.

No router was required for this errno translation. Please review the regmap
callback contract and timeout propagation.

## Package build validation

### DIFF
--- a/package/kernel/rtl837x/src/rtl837x_mdio.c
+++ b/package/kernel/rtl837x/src/rtl837x_mdio.c
@@ -90,8 +90,8 @@ static int rtl837x_mdio_write(void *ctx, u32 reg, u32 val)
 	if (ret < 0)
 		goto out_unlock;
 
-    if (ret & 0x4) {
-		ret = RT_ERR_BUSYWAIT_TIMEOUT;
+	if (ret & 0x4) {
+		ret = -ETIMEDOUT;
 		goto out_unlock;
     }
 
@@ -116,8 +116,8 @@ static int rtl837x_mdio_write(void *ctx, u32 reg, u32 val)
 	if (ret < 0)
 		goto out_unlock;
 
-    if (ret & 0x4) {
-		ret = RT_ERR_BUSYWAIT_TIMEOUT;
+	if (ret & 0x4) {
+		ret = -ETIMEDOUT;
 		goto out_unlock;
     }
 	ret = 0;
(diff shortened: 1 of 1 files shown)


perceival:
**Merging.** `RT_ERR_BUSYWAIT_TIMEOUT` is enum value `0x0a` — **positive** — and it was being returned straight out of `reg_read`/`reg_write`. regmap's contract is 0 or a negative errno, so this is a genuine violation and `-ETIMEDOUT` is the right conversion. All four sites are covered, and nothing else in the transaction or locking sequence changes (`REGCACHE_NONE`, custom lock ops).

One correction to the rationale, so the commit message doesn't overstate it: the description says a positive return "can be interpreted as a successful result", and on the current call chain that isn't true. `_regmap_read()`, `_regmap_write()` and `_regmap_update_bits()` in 6.18.39 all test `ret == 0` / `ret != 0` rather than `ret < 0`, and the only consumer — `rtl8373_asicdrv.c` — does `if (ret) return RT_ERR_SMI;`. So `+10` currently propagates as *an* error and nothing silently succeeds.

That makes this a latent-bug fix rather than a live one, which is still worth merging: the moment any caller uses the conventional `ret < 0` test, or regmap's internals change to, a timeout would read as success. Worth fixing now while it's cheap.

MNeroba:
The PR description has been updated to make the callback contract explicit: vendor-positive statuses are translated to Linux errno before returning from the regmap callbacks, while the existing non-zero error handling at the callers remains unchanged. No unrelated MDIO behavior was added.

PR #58, opened 2026-09-01
rtl837x: isolate FDB entries by tag_8021q database

MNeroba:
## Summary

Translate zero-VLAN FDB operations into the reserved tag_8021q VID for the target DSA database.

- Standalone port databases use `dsa_tag_8021q_standalone_vid()`.
- Bridge databases use `dsa_tag_8021q_bridge_vid()`.
- Unsupported database types return `-EOPNOTSUPP` when no explicit VLAN ID is supplied.

The translated VID is used consistently for both the IVL flag and the hardware FDB key.

## Why this is correct

With tag_8021q, a zero `vid` in a DSA FDB callback identifies the database context rather than a literal hardware VLAN. The RTL837x driver previously programmed all such entries with VID/FID zero, so standalone and bridge entries could share the same hardware lookup domain despite `ds->fdb_isolation` being enabled.

The driver now maps the DSA database context to the same reserved management VID scheme used by tag_8021q. Explicit nonzero VLAN IDs remain unchanged. Add and delete use the same translation, preserving key symmetry, and the existing CPU-port shortcut is retained.

This is intentionally limited to unicast FDB key construction. It does not change VLAN allocation, bridge membership, CPU flooding, multicast handling, or the existing `fdb_isolation` declaration.

## Validation

- `git diff --check`: passed.
- Confirmed DSA `struct dsa_db` supports the standalone-port and bridge database contexts used by this driver.
- Confirmed the tag_8021q standalone and bridge VID helpers are available to the current 6.18 target.
- Confirmed add and delete construct identical translated keys.

No router was required for this key-mapping fix. Please review the database-to-VID mapping and the unsupported database behavior.

## Package build validation

### DIFF
--- a/package/kernel/rtl837x/src/rtl837x_dsa_ops.c
+++ b/package/kernel/rtl837x/src/rtl837x_dsa_ops.c
@@ -800,6 +800,25 @@ static void rtl837x_port_fast_age(struct dsa_switch *ds, int port)
 			ret);
 }
 
+static int rtl837x_fdb_vid(u16 vid, struct dsa_db db, u16 *fdb_vid)
+{
+	if (vid) {
+		*fdb_vid = vid;
+		return 0;
+	}
+
+	switch (db.type) {
+	case DSA_DB_PORT:
+		*fdb_vid = dsa_tag_8021q_standalone_vid(db.dp);
+		return 0;
+	case DSA_DB_BRIDGE:
+		*fdb_vid = dsa_tag_8021q_bridge_vid(db.bridge.num);
+		return 0;
+	default:
+		return -EOPNOTSUPP;
+	}
+}
+
 static int rtl837x_port_vlan_filtering(struct dsa_switch *ds, int port,
 				       bool vlan_filtering,
 				       struct netlink_ext_ack *extack)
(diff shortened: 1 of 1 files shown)


perceival:
**Merging.** This mirrors the canonical upstream implementation almost exactly — `sja1105_fdb_add()` in `drivers/net/dsa/sja1105/sja1105_main.c` uses the same `if (!vid)` guard, the same `DSA_DB_PORT` -> `dsa_tag_8021q_standalone_vid(db.dp)` / `DSA_DB_BRIDGE` -> `dsa_tag_8021q_bridge_vid(db.bridge.num)` mapping, and the same `-EOPNOTSUPP` default. Both helpers are exported for the 6.18 target.

The premise holds too: this driver sets `ds->fdb_isolation = true`, so DSA passes a meaningful `struct dsa_db` and expects the driver to keep databases apart. Programming every zero-vid entry with VID/FID 0 meant standalone and bridge entries shared one hardware lookup domain, which is exactly what `fdb_isolation` promises not to happen. Add and delete construct identical keys, so entries remain deletable.

One consequence worth stating explicitly, since it is a behaviour change rather than a bug: because the translated VID is always nonzero, `l2.ivl` is now always 1, whereas zero-vid entries previously programmed `ivl = 0`. Moving those entries from shared to independent VLAN learning is the intent of the change and is consistent with `fdb_isolation`, but it does shift which hardware lookup resources they consume. Worth a glance on hardware if FDB capacity ever looks tight.

The translated VIDs are guaranteed to exist in the VLAN table, since `tag_8021q` programs them through the driver's own `tag_8021q_vlan_add`, so the keys are meaningful.

PR #59, opened 2026-09-01
rtl837x: reject multiple switch instances

MNeroba:
## Summary

Protect the vendor SDK global private pointer and reject a second RTL837x
switch probe while another instance is active.

The claim and release paths are serialized by a mutex. The claim is made
before hardware initialization. Normal remove releases the global pointer;
shutdown deliberately keeps ownership until remove because associated
debugfs, SFP, GPIO, and DSA state may still reference the switch context.

## Why this is correct

The RTL837x SDK uses global state, including `rtl_gbl_priv`, for register-map
access and mapper operations. The driver previously assigned that pointer
unconditionally during probe, so a second switch instance could silently
redirect SDK accesses to the newer device and leave the first instance using
the wrong context.

The driver now atomically checks and claims the global context, returning
`-EBUSY` when another instance owns it. Cleanup only clears the pointer when
it still belongs to the same switch, preventing one instance from clearing
another instance's ownership. The claim is placed before the first hardware
initialization call. If the claim fails, any acquired master-netdev reference
is released; failures after the claim clear the ownership and retain the
devres-safe probe cleanup from the base branch.

PR #52, which provides that devres-safe failed-probe cleanup, has been merged
into `flint3-be9300`. This branch is rebased onto
`79d086e797e60c0dfb195d4556a5bf5ad1332fc8` and contains only the remaining
singleton ownership guard; the #52 cleanup is inherited from the base and is
not duplicated here.

This is intentionally limited to global-instance ownership. It does not
refactor the vendor SDK globals, change normal DSA registration, or alter
hardware configuration.

## Validation

- `git diff --check`: passed.
- Final diff: 1 file changed, 26 insertions, 1 deletion.
- Confirmed `rtl_gbl_priv` is claimed before `rtl8372n_hw_init()`, the first
  hardware/SDK initialization call.
- Confirmed a second probe returns `-EBUSY` without overwriting the active
  pointer and releases its master-netdev reference.
- Confirmed normal remove clears only its own pointer and
  `rtl837x_mdio_shutdown()` does not clear global ownership.
- Confirmed the final diff contains only singleton ownership logic; the
  failed-probe cleanup is inherited from #52.

## Package build validation

- `make package/kernel/rtl837x/compile V=s`: attempted in the isolated
  worktree on macOS; OpenWrt stopped before package compilation because the
  Apple-provided GNU Make 3.81 is unsupported.
- No successful package-build result is claimed; a Linux case-sensitive
  OpenWrt build host is still required.

### DIFF
--- a/package/kernel/rtl837x/src/rtl837x_mdio.c
+++ b/package/kernel/rtl837x/src/rtl837x_mdio.c
@@ -888,10 +888,28 @@ static const struct of_device_id rtk_gsw_match[] = {
 
 MODULE_DEVICE_TABLE(of, rtk_gsw_match);
 
+static DEFINE_MUTEX(rtl837x_instance_lock);
+
+static int rtl837x_claim_global_priv(struct rtk_gsw *gsw)
+{
+	int ret = 0;
+
+	mutex_lock(&rtl837x_instance_lock);
+	if (rtl_gbl_priv)
+		ret = -EBUSY;
+	else
+		rtl_gbl_priv = gsw;
+	mutex_unlock(&rtl837x_instance_lock);
+
+	return ret;
+}
+
 static void rtl837x_clear_global_priv(struct rtk_gsw *gsw)
 {
+	mutex_lock(&rtl837x_instance_lock);
 	if (rtl_gbl_priv == gsw)
 		rtl_gbl_priv = NULL;
+	mutex_unlock(&rtl837x_instance_lock);
 }
 
 static int rtl837x_dsa_probe(struct mdio_device *mdiodev)
@@ -1029,8 +1047,15 @@ static int rtl837x_dsa_probe(struct mdio_device *mdiodev)
 	dev_info(gsw->dev, "rtl837x dev info:smi-addr:%d configured-cpu-port:%u serdes-mode:%d swap_cfg:0x%x\n",
 						 gsw->mdio_addr, gsw->cpu_port, gsw->sds0mode, *(uint8_t*)&(gsw->swap_cfg));
 
+	ret = rtl837x_claim_global_priv(gsw);
+	if (ret) {
+		dev_err(dev, "another RTL837x switch instance is already active\n");
+		if (master)
+			dev_put(master);
+		return ret;
+	}
+
 	dev_set_drvdata(dev, gsw);
-	rtl_gbl_priv = gsw;
 
 	ret = rtl8372n_hw_init(gsw, gsw->swap_cfg);
 	if (ret)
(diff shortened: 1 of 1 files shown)


perceival:
**The guard is right; please drop the `shutdown` hunk.**

The claim/release logic is sound: mutex-serialised test-and-set, the claim placed before the first SDK-touching call (`rtl8372n_hw_init()`), both existing failure paths still clear the global, the ownership-checked clear stops a second instance from clearing the first's pointer, and the `-EBUSY` path correctly `dev_put()`s the master netdev taken via `of_find_net_device_by_node()`. Process context throughout.

**The problem is `rtl837x_clear_global_priv()` in `rtl837x_mdio_shutdown()`.** Unlike `rtl837x_dsa_remove()`, the shutdown path does not tear down the consumers that reach the SDK through `rtl_gbl_priv`, so clearing the pointer there opens a NULL-deref window and buys nothing:

- `rtl837x_debug_proc_deinit()` is not called, so the debugfs `reg`/`phyreg_mmd`/`sdsreg` files stay live. On `reboot`: mdio shutdown clears the pointer → any surviving process or init script doing `echo "r 0x7b20" > .../reg` → `rtl8373_getAsicReg()` → `regmap_read(rtl_gbl_priv->map, …)` on NULL → oops during reboot.
- `sfp_bus_del_upstream()` is not called, so an SFP insert/remove after the clear runs `rtl837x_sfp_module_insert()` → `rtk_sdsMode_set()` → same deref.
- The devm gpiochip (registered when `gpio-controller` is in DT) isn't removed at shutdown either, so `rtl837x_gpio_set()` → `gsw->pMapper->…` → same deref.

Nothing re-probes after `.shutdown` — rebind goes through `.remove` — so there's no upside to releasing the singleton there. Either drop that hunk or move it after explicit debugfs/SFP/GPIO teardown.

(Unrelated to this PR, but noted while looking: `remove()` has the same ordering hazard already, clearing the pointer before devres unwinds the gpiochip.)

**Merge ordering:** the `-EBUSY` path relies on devres to free `gsw` while the neighbouring paths still `devm_kfree()` explicitly. That's consistent only once #52 lands, as your description says — please merge #52 first or rebase.

MNeroba:
Pushed `406dd80ff7` to remove `rtl837x_clear_global_priv()` from the MDIO shutdown path. Normal device removal still releases the global pointer; shutdown now leaves it available while debugfs, SFP, and GPIO consumers can still reference the device. The description was updated to state this lifecycle distinction.

PR #60, opened 2026-09-01
rtl837x: offload bridge flood flags

MNeroba:
## Summary

Extend the RTL8373 DSA bridge-port flag support with the three Linux flood
flags:

- `BR_FLOOD` through the unknown-unicast flood mask;
- `BR_MCAST_FLOOD` through the unknown L2, IPv4 multicast, and IPv6 multicast
  masks;
- `BR_BCAST_FLOOD` through the broadcast flood mask.

The learning, hairpin, isolation, and FDB functionality used by the shared
bridge-flag dispatcher is already in `flint3-be9300`; this PR adds only the
flood-mask operations.

## Implementation

The DSA callback uses the existing RTL8373 RTK/DAL
`rtk_l2_floodPortMask_{get,set}()` operations. It does not access registers
directly. Each update reads the complete logical port mask, changes only the
requested user-port bit, and preserves the CPU and all other port bits.

All required masks are read before any write. A switch-level mutex serializes
updates, and a failed write restores every mask already attempted, including
the failing operation. Unsupported flags return `-EOPNOTSUPP` with the
existing extack message; invalid or non-user ports retain `-EINVAL`.

## Scope

This PR does not change `tag_8021q`, VLAN/PVID, STP, LAG, MDB, or the merged
learning/hairpin/isolation implementations. It does not touch the CPU-port
bit in any flood mask.

## Validation

- Rebased onto current `flint3-be9300` at
  `a655214e41fb8715d87e923e6a31bb1e9c6508ba`.
- Final diff is limited to `rtl837x_common.h`, `rtl837x_dsa_ops.c`, and
  `rtl837x_mdio.c`.
- `git diff --check`: passed.
- `scripts/checkpatch.pl --no-tree --terse`: passed with no new errors or
  warnings.
- OpenWrt package compilation was not completed on this macOS host because
  the repository prerequisite checks require a case-sensitive filesystem and
  GNU host tools.

## Dependencies

The earlier bridge-flag prerequisites (#63, #64, and #65), the FDB mapper
(#71), the link-down/FDB age work (#69), the regression coverage (#70), and
VLAN-specific fast ageing (#72) are already in the base branch. No open code
dependency remains, and this branch contains no duplicate implementation from
those changes.

Hardware review should still exercise each flood flag, preservation of
unrelated mask members including the CPU port, and rollback after an injected
RTK/MDIO failure.

### DIFF
--- a/package/kernel/rtl837x/src/rtl837x_common.h
+++ b/package/kernel/rtl837x/src/rtl837x_common.h
@@ -98,6 +98,7 @@ struct rtk_gsw {
 	struct regmap		*map;
 	struct regmap		*map_nolock;
 	struct mutex		map_lock;
+	struct mutex		flood_lock;
 
 	struct gpio_desc *reset_pin;
 	int mdio_addr;
--- a/package/kernel/rtl837x/src/rtl837x_dsa_ops.c
+++ b/package/kernel/rtl837x/src/rtl837x_dsa_ops.c
@@ -74,7 +74,109 @@ static int rtl837x_set_hairpin(struct rtk_gsw *gsw, int port, bool enable)
 }
 
 #define RTL837X_SUPPORTED_BRIDGE_FLAGS \
-	(BR_LEARNING | BR_HAIRPIN_MODE | BR_ISOLATED)
+	(BR_LEARNING | BR_HAIRPIN_MODE | BR_ISOLATED | BR_FLOOD | \
+	 BR_MCAST_FLOOD | BR_BCAST_FLOOD)
+
+struct rtl837x_flood_mask_update {
+	rtk_l2_flood_type_t type;
+	rtk_portmask_t old;
+	rtk_portmask_t new;
+	bool valid;
+	bool attempted;
+};
+
+static int rtl837x_flood_mask_update_prepare(
+		struct rtl837x_flood_mask_update *update, int port, bool enable)
+{
+	rtk_api_ret_t ret;
+
+	ret = rtk_l2_floodPortMask_get(update->type, &update->old);
+	if (ret != RT_ERR_OK)
+		return rtl837x_to_errno(ret);
+
+	update->new = update->old;
+	if (enable)
+		RTK_PORTMASK_PORT_SET(update->new, port);
+	else
+		RTK_PORTMASK_PORT_CLEAR(update->new, port);
+
+	update->valid = true;
+	return 0;
+}
+
+static void rtl837x_flood_mask_rollback(struct rtk_gsw *gsw,
+					struct rtl837x_flood_mask_update *updates,
+					unsigned int count)
+{
+	unsigned int i;
+	rtk_api_ret_t ret;
+
+	for (i = 0; i < count; i++) {
+		if (!updates[i].valid || !updates[i].attempted)
+			continue;
+
+		ret = rtk_l2_floodPortMask_set(updates[i].type,
+						       &updates[i].old);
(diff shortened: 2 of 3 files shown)


MNeroba:
Source-level audit for reviewers:

- Linux 6.18 exposes `port_pre_bridge_flags` and `port_bridge_flags` with `struct switchdev_brport_flags`; the pre-callback rejects every changed flag outside `BR_FLOOD`, `BR_MCAST_FLOOD`, and `BR_BCAST_FLOOD`.
- The existing RTL8373 RTK/DAL API provides get/set operations for `FLOOD_UNKNOWNDA`, `FLOOD_UNKNOWNL2MC`, `FLOOD_UNKNOWNV4MC`, `FLOOD_UNKNOWNV6MC`, and `FLOOD_BC`, so the DSA callback does not need raw register access.
- Each operation reads the complete mask, changes only the requested logical user-port bit, and preserves the CPU and other-port bits. Multicast flood is applied consistently to the unknown L2, IPv4, and IPv6 multicast domains.
- All reads complete before writes. A switch-level mutex serializes updates; a failed write restores all attempted masks, including the failed operation, and reports rollback failures while returning the original RTK error.

This is intentionally independent of #3/#58 and does not touch tag_8021q, VLAN/PVID, STP, LAG, or MDB state. Static checks passed (`git diff --check` and `checkpatch`); the requested profile build and bench hardware validation remain pending and are not being claimed as completed.

perceival:
**Merging with one nit.** I checked the things most likely to be wrong here and they're right.

**Port namespace is correct**, which was my main concern. `rtk_l2_floodPortMask_{get,set}` operate in the logical domain — the getter ends with `rtk_switch_portmask_P2L_get()` and the setter begins with `rtk_switch_portmask_L2P_get()` — and the RTL8373 HAL's L2P/P2L tables are the identity map for ports 0-9, which matches how the rest of the driver passes DSA port indices straight to RTK APIs. The read-modify-write loses no bits either: `valid_portmask` covers all ten ports, so the P2L→L2P round trip is lossless and the CPU-port bit (and any vendor-set bit) survives.

**Return-value contract is fine.** `dsa_port_inherit_brport_flags()` and `dsa_port_clear_brport_flags()` iterate flags one bit at a time and both tolerate `-EOPNOTSUPP`, so rejecting `BR_LEARNING`/`BR_PORT_LOCKED` doesn't break bridge join or leave. Adding these callbacks also doesn't flip `dsa_port_can_configure_learning()` to true, so STP fast-age behaviour and `dsa_port_supports_mst()` are unchanged.

**No tag_8021q interaction**, which is worth stating given how much else on this switch does interact with it: flood masks are pure egress port masks, orthogonal to VID-based forwarding, and the CPU-port bit is never touched — so host-bound unknown unicast keeps flooding to the CPU, which `rtl837x_port_fdb_add`'s comment depends on.

Rollback logic is sound: `attempted` is set before the write, so a partially-applied failing write is restored too, and entries where old == new are skipped and correctly left un-attempted. `flood_lock` is sufficient (no other flood-mask user) and is initialised on the only probe path.

Nit: both callbacks return `-EINVAL` for a non-user port. `dsa_port_inherit_brport_flags()` aborts the bridge join on any error that isn't `-EOPNOTSUPP`, so if a future kernel ever probed these on a CPU/DSA port, `ip link set lan1 master br-lan` would fail outright rather than degrade. `-EOPNOTSUPP` is the safer value. Unreachable in 6.18.39 — all call sites are user ports — so not blocking.

PR #61, opened 2026-09-01
rtl837x: document MTU and jumbo-frame research

MNeroba:
## Summary

Document the RTL8373 MTU/jumbo-frame audit and the evidence required before a
DSA MTU callback can be implemented. This is a research-only PR: it adds no
MTU callback and no register programming.

## Findings

- The RTL8373 DAL exposes per-port RX max-length set/get operations with two
  speed buckets: FE/10M/100M and 1G/2.5G/5G/10G.
- The RX register contains two independent 14-bit fields, one per speed
  bucket. Generated definitions also contain a separate TX max-length
  register, but no corresponding RTL8373 TX DAL mapper operation was found.
- The SDK's tag-length adjustment API does not define how its accounting
  relates to the `tag_8021q` transport used by this DSA driver.
- Generic RTK max-packet-length profile mapper entries are disabled for
  RTL8373, and no RTL8373 profile implementation was found.
- Public RTL8372/3 firmware confirms per-port RX register use and the
  64..16383 range, but does not establish Linux DSA, TX, CPU-port, tag, FCS,
  or rollback semantics.
- Because this driver does not implement `port_max_mtu`, generic DSA keeps
  user ports at the ordinary `ETH_DATA_LEN` (1500-byte) limit. A larger
  bootloader-configured RX value does not by itself enable jumbo frames via
  DSA.

## Decision

Do not implement `port_change_mtu` or `port_max_mtu` until hardware readback
and traffic tests establish whether Linux MTU controls RX, TX, or both; how
the CPU port is handled; and the exact VLAN/FCS/`tag_8021q` conversion. The
note records the required bench matrix and follow-up questions.

## Validation

- Rebased onto current `flint3-be9300` at
  `a655214e41fb8715d87e923e6a31bb1e9c6508ba`.
- Documentation-only final diff: `package/kernel/rtl837x/MTU-JUMBO-RESEARCH.md`.
- `git diff --check`: passed.
- No target package build or hardware validation was performed.

## References

- [RTL8373 port max-length DAL](https://github.com/perceival/openwrt-flint3/blob/flint3-be9300/package/kernel/rtl837x/src/rtk-api/dal/rtl8373/dal_rtl8373_port.c)
- [RTL8373 generated register definitions](https://github.com/perceival/openwrt-flint3/blob/flint3-be9300/package/kernel/rtl837x/src/rtk-api/dal/rtl8373/rtl8373_reg_definition.h)
- [Linux RTL8365MB DSA MTU implementation](https://github.com/torvalds/linux/blob/master/drivers/net/dsa/realtek/rtl8365mb_main.c)
- [Related RTL8372/3 firmware research](https://github.com/logicog/RTLPlayground)

### DIFF
--- a/package/kernel/rtl837x/MTU-JUMBO-RESEARCH.md
+++ b/package/kernel/rtl837x/MTU-JUMBO-RESEARCH.md
@@ -1,0 +1,81 @@
+# RTL8373 MTU and jumbo-frame offload research
+
+Status: research-only. This note deliberately does not add a DSA MTU callback
+or change any RTL8373 register.
+
+## Question
+
+Can the RTL837x DSA driver implement `port_change_mtu` and
+`port_max_mtu` with an exact Linux-to-hardware frame-length mapping for
+ordinary and jumbo frames?
+
+## Evidence in the current RTL8373 SDK
+
+The chip-specific DAL provides `dal_rtl8373_portMaxLen_set()` and
+`dal_rtl8373_portMaxLen_get()`. Their own comments describe these as *port RX
+max length* operations. They select two speed buckets: FE/10M/100M and
+1G/2.5G/5G/10G. The generated register definitions expose a 14-bit RX field at
+`MAC_L2_PORT_MAX_LEN_CTRL_ADDR(port)` and a separate per-port TX max-length
+register, but no RTL8373 DAL set/get API for the TX register was found.
+
+The same DAL file has `portMaxLenIncTag_{set,get}()` for a generic tag-length
+adjustment. The SDK does not establish whether that bit accounts for the
+`tag_8021q` transport used by this DSA driver, which tags are included, or how
+the adjustment interacts with Linux's MTU and FCS accounting.
+
+The generic RTK API declares profile-style max-packet-length operations
+(`rtk_switch_portMaxPktLen_{set,get}()` and
+`rtk_switch_maxPktLenCfg_{set,get}()`), but the RTL8373 mapper leaves those
+entries disabled and no RTL8373 profile implementation was found. Therefore
+the generic API cannot currently be used as a proven DSA MTU abstraction.
+
+## Other repository evidence
+
+The public [RTLPlayground RTL8372/3 firmware](https://github.com/logicog/RTLPlayground)
+reads and writes `0x1250 + (port << 8)`, masks the 14-bit RX value, accepts
+64..16383, and presents example choices such as 1522, 1536, 1552, 9216, and
+16383. This confirms that a per-port RX frame-length control is usable on
+related RTL837x hardware. It does not document DSA MTU semantics, the TX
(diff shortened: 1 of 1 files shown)


MNeroba:
Research result for reviewers:

- `dal_rtl8373_portMaxLen_{set,get}()` is explicitly an RX max-length API with FE/10M/100M and 1G/2.5G/5G/10G selectors.
- The generated RTL8373 definitions contain a separate TX max-length register, but the current SDK has no TX DAL/RTK mapper operation.
- `portMaxLenIncTag_{set,get}()` exists, yet its relationship to the DSA `tag_8021q` transport, VLAN headers, and FCS is not defined by the SDK.
- The generic profile max-packet-length mapper is disabled for RTL8373. The public RTLPlayground firmware confirms RX register read/write and the 64..16383 range, but does not establish Linux DSA ownership or TX behavior.

Accordingly this PR records the blocker and a bench-test matrix, without adding a callback that could advertise an MTU while programming only one half of the frame path.

perceival:
**Merging with two small corrections.** I verified every hardware claim against the tree rather than the prose, and they hold:

- `dal_rtl8373_portMaxLen_{set,get}(port, type, len)` exist and select two speed buckets via `type` (0 = 10M/100M, 1 = 1G and above). ✅
- Per-port 14-bit field: `RTL8373_MAC_L2_PORT_MAX_LEN_CTRL_ADDR(port) = 0x1250 + (port << 8)`, masks `0x3FFF` — and 16383 is exactly the 14-bit maximum, matching the RTLPlayground figures. ✅
- A separate TX register with no DAL accessor: `0x4E90 + (port << 2)` exists and nothing in `dal/rtl8373/` touches it. ✅
- `portMaxLenIncTag_{set,get}` drive bit 28 of the same RX register. ✅
- The generic profile mapper entries are commented out for RTL8373. ✅
- The RTL8365MB comparison is accurate — `rtl8365mb_port_change_mtu()` programs the *global* max-length register from the CPU port only, using `new_mtu + VLAN_ETH_HLEN + ETH_FCS_LEN`. ✅
- The driver genuinely has no `port_change_mtu`/`port_max_mtu` and never calls `portMaxLen*`, so the "no MTU control today" premise holds. ✅

Corrections:
- The RTL8365MB reference URL points at `942-04-7.2-net-dsa-realtek-rtl8365mb-prepare-for-multiple-sourc.patch`, which is the multi-source refactor, not the MTU code being described. Cite `drivers/net/dsa/realtek/rtl8365mb_main.c`.
- "a 14-bit RX field" is really *two* 14-bit fields in one register (bits 0-13 and 14-27), one per speed bucket. The DAL paragraph already says "two speed buckets", so the note isn't wrong overall — just the register sentence.

Worth adding, since it's the practical consequence a reader will want: with no `port_max_mtu`, DSA caps user ports at `ETH_DATA_LEN` (1500) regardless of what the bootloader left in the hardware RX register.

PR #62, opened 2026-09-01
rtl837x: add DSA FDB dump support

MNeroba:
## Summary

Add focused DSA `port_fdb_dump` support for RTL837x so Linux can enumerate
hardware-learned unicast entries for a switch port. This is the independently
reviewable FDB part extracted from #3; MDB and multicast behavior remain out
of scope.

## Implementation

- Add `.port_fdb_dump` using `READMETHOD_NEXT_L2UCSPA`.
- Stop iteration only on `RT_ERR_L2_ENTRY_NOTFOUND`; propagate every other
  SDK/DAL failure through `rtl837x_to_errno()`.
- Skip entries returned for a different source port and multicast MACs before
  invoking the DSA callback.
- Preserve explicit IVL VLAN IDs and report the driver's internal
  `tag_8021q` transport VIDs as Linux VID 0.
- Remove the two unconditional vendor console prints from the
  `NEXT_L2UCSPA` DAL path.
- Propagate callback errors such as `-EMSGSIZE`.

## Integration

The RTL8373 `l2_addr_next_get` mapper is already provided by merged #71 and is
inherited from `flint3-be9300`; this PR does not duplicate that mapper change.
The branch is rebased onto the current `flint3-be9300` tip
`ccf16a0878d1985210618fa7b2f7905427fbd68d`; its new head is
`2de1c6fa3a398a301d73b04703a18290420849b5`.
The final diff contains only the iterator cleanup and DSA dump callback.

## Scope

This PR intentionally does not add MDB add/delete, IGMP/MLD state changes,
multicast lookup-mode changes, multicast CPU membership, flood control, or
LAG-specific FDB handling.

## Validation

- Final diff is limited to
  `package/kernel/rtl837x/src/rtk-api/dal/rtl8373/dal_rtl8373_lut.c` and
  `package/kernel/rtl837x/src/rtl837x_dsa_ops.c`.
- `git diff --check`: passed.
- `scripts/checkpatch.pl --no-tree --terse`: passed with no new errors or
  warnings.
- No OpenWrt package build was completed on this macOS host because the
  repository prerequisite checks require a case-sensitive filesystem and GNU
  host tools.

Suggested hardware checks are dynamic/static FDB reporting, source-port
separation, multicast exclusion, VLAN-aware VID preservation,
`tag_8021q`/VID-0 normalization, callback error propagation, and aged-entry
removal.

The maintainer review follow-up is now in the branch; please re-review the
updated FDB-only implementation. PR #3 remains the separate MDB/reference
work and is not a dependency of this change.

### DIFF
--- a/package/kernel/rtl837x/src/rtk-api/dal/rtl8373/dal_rtl8373_lut.c
+++ b/package/kernel/rtl837x/src/rtk-api/dal/rtl8373/dal_rtl8373_lut.c
@@ -206,7 +206,6 @@ static rtk_api_ret_t _rtl8373_getL2LookupTb(rtk_uint32 method, rtl8373_luttb *pL
             retVal = rtl8373_setAsicRegBits(RTL8373_ITA_CTRL0_ADDR, RTL8373_ITA_CTRL0_TBL_ADDR_MASK, pL2Table->address);
             if(retVal != RT_ERR_OK)
                 return retVal;
-            rtlglue_printf("spa is %d\n",pL2Table->spa);
             retVal = rtl8373_setAsicRegBits(RTL8373_ITA_L2_CTRL_ADDR, RTL8373_ITA_L2_CTRL_PORT_NUM_MASK, pL2Table->spa);
             if(retVal != RT_ERR_OK)
                 return retVal;
@@ -215,7 +214,6 @@ static rtk_api_ret_t _rtl8373_getL2LookupTb(rtk_uint32 method, rtl8373_luttb *pL
             retVal = rtl8373_setAsicReg(RTL8373_ITA_CTRL0_ADDR, tblCmd);
             if(retVal != RT_ERR_OK)
                 return retVal;
-            rtlglue_printf("tblcmd is 0x%x\n",tblCmd);
             break;
         default:
             return RT_ERR_INPUT;
--- a/package/kernel/rtl837x/src/rtl837x_dsa_ops.c
+++ b/package/kernel/rtl837x/src/rtl837x_dsa_ops.c
@@ -1929,6 +1929,46 @@ static int rtl837x_port_fdb_del(struct dsa_switch *ds, int port,
 	return rtl837x_to_errno(ret);
 }
 
+static int rtl837x_port_fdb_dump(struct dsa_switch *ds, int port,
+				 dsa_fdb_dump_cb_t *cb, void *data)
+{
+	struct rtk_gsw *gsw = ds->priv;
+	u32 max = RTK_MAX_LUT_ADDRESS;
+	u32 address = 0;
+	int ret;
+
+	if (!rtl837x_valid_port(gsw, port))
+		return -EINVAL;
+
+	while (address < max) {
+		rtk_l2_ucastAddr_t l2 = { 0 };
+		u16 vid;
+
+		ret = rtk_l2_addr_next_get(READMETHOD_NEXT_L2UCSPA, port,
+					   &address, &l2);
+		if (ret == RT_ERR_L2_ENTRY_NOTFOUND ||
+		    ret == RT_ERR_L2_L2UNI_PARAM)
+			break;
+		if (ret != RT_ERR_OK)
+			return rtl837x_to_errno(ret);
+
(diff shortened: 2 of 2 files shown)


perceival:
**Needs changes — one blocker.** The DSA contract itself is right: the `dsa_fdb_dump_cb_t` signature matches, propagating the callback's `-EMSGSIZE` is required and handled, `.l2_addr_next_get` is wired in the RTL8373 mapper, and the loop bound is sound (`RTK_MAX_LUT_ADDRESS` is 4160, the DAL rejects `> 4159`, so `while (address < 4160)` can't overrun). The tag_8021q normalisation is right too — `vid_is_dsa_8021q()` tests the RSV nibble, which is set on both the standalone per-port VIDs and `dsa_tag_8021q_bridge_vid()`, so both collapse to VID 0 while a VLAN-aware bridge's real VIDs pass through. And `l2.ivl ? l2.vid_fid : 0` round-trips correctly against `rtl837x_port_fdb_add`'s `l2.ivl = vid ? 1 : 0`.

**1. (blocking) Every dumped entry emits two unconditional `printk`s, with rtnl held.**

The `NEXT_L2UCSPA` branch of `_rtl8373_getL2LookupTb()` — the exact path this dump uses — contains leftover vendor debug prints:

```
dal_rtl8373_lut.c:209:   rtlglue_printf("spa is %d\n", pL2Table->spa);
dal_rtl8373_lut.c:218:   rtlglue_printf("tblcmd is 0x%x\n", tblCmd);
```

and `rtk_types.h:111` is `#define rtlglue_printf printk` under `__KERNEL__` — no level, so KERN_DEFAULT, so console.

Failure scenario: a deployed AP with ~400 learned MACs. `bridge fdb show` — or anything polling it, LuCI status included — walks all four LAN ports and emits ~800 console lines per invocation, while `rtnl_lock` is held for the entire walk because `ndo_fdb_dump` runs under rtnl. On a 115200 serial console that's tens of seconds with rtnl blocked: netifd, hostapd and `ip` all stall, and it will trip the hung-task/soft-lockup detectors. Please delete or comment those two lines — the same file already has four other prints commented out exactly that way (lines 78/426/722).

**2. No check that the returned entry belongs to `port`, or that it's unicast.**

The vendor marked this function `//not use, use dal_rtl8373_l2_entry_getNext` immediately above `dal_rtl8373_l2_addr_next_get()`, so the SPA-filtered search is untested vendor code — and your own description lists "entry learned on lan1 is not dumped for lan2" as an unrun hardware check. Compounding it, `_rtl8373_fdbStSmi2User()` only populates `spa` and `nosalearn` in the auto-learning *unicast* branch; for an L2-multicast or L3 hit both are left at 0. So if the SPA filter turns out to be soft, or a multicast entry lands in the result, it is reported with `spa = 0` and `is_static = false` — a multicast MAC appears in `bridge fdb show dev lan1`, and on a switch where port 0 exists you'd get every port's MACs attributed to lan1.

Two cheap guards make that self-correcting instead of a hardware-dependent mystery:

```c
if (l2.port != port)
        continue;      /* SPA filter didn't hold */
if (is_multicast_ether_addr(l2.mac.octet))
        continue;      /* MC/L3 hit: spa/is_static are invalid */
```

(with `address++` before the `continue`). Note an L2-multicast hit would also fire `rtlglue_printf("l2 multicast\n")` at line 91, compounding finding 1.

**3. (non-blocking) A mid-dump SDK failure aborts the whole `bridge fdb show`.** A transient `RT_ERR_BUSYWAIT_TIMEOUT` from the poll loop becomes `-EIO` and kills the dump for *all* netdevs, not just this port. The index-resumed netlink dump already tolerates entries aging out between chunks, so a transient MDIO timeout is worth softening rather than hard-failing.

Also worth knowing there's a competing FDB-dump implementation in #3 from another contributor; whichever lands, the two shouldn't both go in.

MNeroba:
Pushed `feea606dc5` to finish the FDB-dump cleanup: the two unconditional DAL console prints in the `NEXT_L2UCSPA` path are gone, and dump results for another port or multicast addresses are skipped before calling the DSA callback. SDK errors still propagate. The body and test notes were updated.

PR #63, opened 2026-09-02
rtl837x: offload bridge port learning flag

MNeroba:
## Summary

Offload the Linux `BR_LEARNING` bridge-port flag on RTL8373 user ports.

## Details

- validate the bridge flag mask before touching hardware;
- return `-EOPNOTSUPP` with an extack message for unsupported flags;
- keep `-EINVAL` for an invalid or non-user port;
- map learning on to the RTL8373 maximum per-port learning limit and learning off to zero, which the DAL documents as disabling learning;
- initialize every standalone user port with hardware learning disabled during switch setup, as expected by Linux DSA.

This uses the existing RTK/DAL learning-limit API and does not add support for
`BR_PORT_LOCKED` or any other bridge flag. The implementation preserves the
current STP/port-enable lifecycle from the base branch. It is intentionally the
first bridge-flag layer; later bridge-flag PRs must extend these callbacks
rather than install competing callback implementations.

## Validation

- Rebased onto current `upstream/flint3-be9300` at
- Final diff is limited to BR_LEARNING helper/callback/setup initialization in
  `rtl837x_dsa_ops.c`.
- `git diff --check`: passed.
  current macOS host because the OpenWrt prerequisite check rejects the
  Apple-provided GNU Make 3.81 before package compilation.
- No hardware test was run.

### DIFF
--- a/package/kernel/rtl837x/src/rtl837x_dsa_ops.c
+++ b/package/kernel/rtl837x/src/rtl837x_dsa_ops.c
@@ -46,6 +46,43 @@ static u32 rtl837x_user_ports(struct rtk_gsw *gsw)
 	return gsw->valid_port_mask & ~BIT(gsw->cpu_port);
 }
 
+#define RTL837X_SUPPORTED_BRIDGE_FLAGS BR_LEARNING
+
+static int rtl837x_port_pre_bridge_flags(struct dsa_switch *ds, int port,
+					 struct switchdev_brport_flags flags,
+					 struct netlink_ext_ack *extack)
+{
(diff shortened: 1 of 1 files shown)


MNeroba:
Publication note: I rechecked this against the Linux 6.18 DSA bridge-flag flow and the RTL8373 learning-limit DAL. Unsupported mask bits return `-EOPNOTSUPP` so normal flag inheritance can continue, while invalid/non-user ports retain `-EINVAL`. The patch is intentionally limited to `BR_LEARNING`; the shared bridge-flag callback must be combined with the related flag patches if they are merged together.

PR #64, opened 2026-09-02
rtl837x: offload bridge hairpin mode

MNeroba:
## Summary

Offload the Linux `BR_HAIRPIN_MODE` bridge-port flag through the RTL8373 source-port permit DAL operation.

## Details

- build on the shared bridge-flag callbacks introduced by merged PR #63;
- support the combined `BR_LEARNING | BR_HAIRPIN_MODE` mask without replacing the merged learning implementation;
- dispatch each supported flag only when it is present in `flags.mask`;
- map hairpin on/off to the documented same-source-and-destination-port permit setting;
- keep `-EOPNOTSUPP` with an extack message for unsupported flags and `-EINVAL` for invalid or non-user ports;
- explicitly disable hairpin for user ports during setup and when a port joins or leaves a bridge;
- roll back setup changes if a later port write fails.

The bridge join/leave wrappers retain the existing `tag_8021q` operations and only add the per-port hairpin lifecycle handling. The CPU port is never changed by the setup defaulting or flag helper.

PR #63 is now inherited from `flint3-be9300`; the next bridge-flag extension (#65) should build on this combined dispatcher rather than add competing callback slots.

## Validation

- Rebased onto current `flint3-be9300` (`9b5702112eead22aa4f5b69001db67e411eea15c`).
- Final PR diff is limited to `package/kernel/rtl837x/src/rtl837x_dsa_ops.c` and hairpin behavior.
- `git diff --check`: passed.
- Full `make package/kernel/rtl837x/compile V=s` was not available on the current macOS host because the OpenWrt prerequisite check cannot be satisfied there.
- No hardware test was run.

### DIFF
--- a/package/kernel/rtl837x/src/rtl837x_dsa_ops.c
+++ b/package/kernel/rtl837x/src/rtl837x_dsa_ops.c
@@ -60,7 +60,21 @@ static int rtl837x_set_learning(struct rtk_gsw *gsw, int port, bool enable)
 	return rtl837x_to_errno(ret);
 }
 
-#define RTL837X_SUPPORTED_BRIDGE_FLAGS BR_LEARNING
+static int rtl837x_set_hairpin(struct rtk_gsw *gsw, int port, bool enable)
+{
+	rtk_api_ret_t ret;
+
+	if (!rtl837x_user_port(gsw, port))
+		return -EINVAL;
+
+	ret = rtk_l2_localPktPermit_set(port,
+						enable ? ENABLED : DISABLED);
+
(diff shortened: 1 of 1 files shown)


MNeroba:
Publication note: I rechecked the hairpin mapping and lifecycle around `tag_8021q` join/leave. The DAL operation is the documented same-source/destination permit control, setup changes have rollback, and unsupported flags return `-EOPNOTSUPP`. This remains limited to `BR_HAIRPIN_MODE`; the callback must be combined with the related bridge-flag patches if they are merged together.

PR #65, opened 2026-09-02
rtl837x: offload bridge port isolation

MNeroba:
## Summary

Add RTL8373 DSA offload for the bridge `BR_ISOLATED` flag while preserving the
already merged learning and hairpin handlers and the existing `tag_8021q`
VLAN forwarding model.

## Details

- maintain bridge membership and the isolated-port mask in driver state under
  a mutex;
- block forwarding only between isolated ports belonging to the same bridge;
- leave CPU reachability and isolated-to-normal-port forwarding available
  through the existing VLAN topology;
- rebuild the hardware isolation matrix on flag changes and bridge join/leave;
- roll back every hardware write already attempted if a matrix update fails;
- stage join/leave state around the existing `tag_8021q` bridge operations;
- preserve the Linux DSA hairpin reset on bridge join and leave.

The shared bridge-flag dispatcher now accepts `BR_LEARNING`,
`BR_HAIRPIN_MODE`, and `BR_ISOLATED`. Learning and hairpin programming are
inherited from merged PRs #63 and #64; this branch adds only the isolation
state, matrix programming, and the required combined join/leave handling.
Unsupported bridge flags return `-EOPNOTSUPP` with an extack message; invalid
or non-user ports still return `-EINVAL`.

If the isolation update fails after `dsa_tag_8021q_bridge_join()` succeeds,
the callback explicitly requests `dsa_tag_8021q_bridge_leave()` and preserves
the original isolation error. The bridge-membership shadow state is committed
only after the matrix update succeeds. The CPU port is never marked isolated
by the bridge-port flag path.

## Current-base integration

The branch is rebased directly onto the current `flint3-be9300` tip
`7c9ce344e23856cf59d49879d6e5238641af050c`, which already contains #63
(learning), #64 (hairpin), #66 (cached `get_stats64`), and #67 (standard
ethtool statistics). The obsolete `port_enable`/`port_disable` and
`port_enabled[]` code from the old pre-#53 base was intentionally not carried
into this PR.

## Validation

- `git diff --check`: passed.
- `scripts/checkpatch.pl --no-tree --terse`: passed with no output/errors for
  the final diff.
- Final diff: two RTL837x source files, isolation fields and the shared
  dispatcher/join/leave integration only.
- Re-read the implementation against Linux 6.18 DSA bridge-flag and
  `tag_8021q` join/leave ordering.
- A full `make package/kernel/rtl837x/compile V=s` is not available on this
  macOS host because OpenWrt rejects Apple's GNU Make 3.81 before package
  compilation.
- No hardware test was run; forwarding and rollback remain bench-validation
  items.

### DIFF
--- a/package/kernel/rtl837x/src/rtl837x_common.h
+++ b/package/kernel/rtl837x/src/rtl837x_common.h
@@ -8,6 +8,7 @@
 
 #include <linux/of_mdio.h>
 #include <linux/netdevice.h>
+#include <linux/mutex.h>
 #include <linux/regmap.h>
 #include <linux/spinlock.h>
 #include <linux/workqueue.h>
@@ -122,6 +123,9 @@ struct rtk_gsw {
 	bool dsa_registered;
 	struct dsa_switch ds;
 	struct net_device *bridge_dev[RTK_MAX_NUM_OF_PORT];
+	/* Hardware-indexed mask of ports with BR_ISOLATED enabled. */
+	u32 isolated_port_mask;
+	struct mutex isolation_lock;
 	struct net_device *ethernet_master;
 	struct sfp_bus *sfp_bus;
 
--- a/package/kernel/rtl837x/src/rtl837x_dsa_ops.c
+++ b/package/kernel/rtl837x/src/rtl837x_dsa_ops.c
@@ -74,7 +74,12 @@ static int rtl837x_set_hairpin(struct rtk_gsw *gsw, int port, bool enable)
 }
 
 #define RTL837X_SUPPORTED_BRIDGE_FLAGS \
-	(BR_LEARNING | BR_HAIRPIN_MODE)
+	(BR_LEARNING | BR_HAIRPIN_MODE | BR_ISOLATED)
+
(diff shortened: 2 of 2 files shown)


MNeroba:
Publication note: I rechecked the isolation matrix against the DSA join/leave ordering. The temporary membership state is used around `tag_8021q`, updates are mutex-protected, and attempted hardware writes are rolled back on failure. The implementation intentionally returns `-EOPNOTSUPP` for unsupported flags and still needs bench validation for isolated-to-isolated, CPU, and isolated-to-normal forwarding.

MNeroba:
Review follow-up: fixed the partial-join path. If `dsa_tag_8021q_bridge_join()` succeeds but the isolation matrix update fails, the callback now calls `dsa_tag_8021q_bridge_leave()` explicitly, leaves both shadow fields unchanged, and returns the original isolation error. The tag rollback API is void, so the code logs that the rollback was requested without replacing the original failure. The leave path remains best-effort and symmetric because its callback cannot return an error.

PR #66, opened 2026-09-02
rtl837x: expose hardware port counters through get_stats64

MNeroba:
## Summary

Expose RTL8373 hardware port counters through DSA `get_stats64` without reading the switch from the network statistics callback.

## Details

- poll the hardware counters once per second after DSA/tag registration;
- cache the resulting `rtnl_link_stats64` values behind a per-port spinlock;
- combine the RTL8373 64-bit packet/octet counters and 32-bit discard/collision counters into standard link statistics;
- detect counter resets before accumulating deltas;
- stop and synchronously drain the delayed work during teardown and shutdown.

The 64-bit octet and packet counters are the reset indicators. A backwards
64-bit value establishes a new hardware baseline but never clears accumulated
Linux `rtnl_link_stats64`. The known 32-bit discard and collision counters use
unsigned modulo-2^32 deltas, so an ordinary wrap is counted instead of being
mistaken for a reset. A failed hardware read leaves both the previous snapshot
and the accumulated statistics untouched for the next poll.

The implementation uses the existing RTK/DAL MIB API; it does not add raw register accesses to the DSA callbacks.

PR #67 is now merged into `flint3-be9300` and is inherited by this branch. The
final PR diff contains only the cached `get_stats64` implementation and its
per-device statistics state; it does not duplicate the structured ethtool
callbacks from #67.

## Validation

- Rebased onto current `upstream/flint3-be9300` at
  `9b5702112eead22aa4f5b69001db67e411eea15c`.
- Final diff is limited to the cached `get_stats64` implementation and its
  per-device statistics state.
- `git diff --check`: passed.
- Corrected the counter references to the RTL8373 API's paired `_H` MIB identifiers.
- Verified that the reset path is baseline-only and that 32-bit counters use
  modulo subtraction without resetting userspace-visible statistics.
  current macOS host because the OpenWrt prerequisite check rejects the
  Apple-provided GNU Make 3.81 before package compilation.
- No hardware test was run.

### DIFF
--- a/package/kernel/rtl837x/src/rtl837x_common.h
+++ b/package/kernel/rtl837x/src/rtl837x_common.h
@@ -7,7 +7,9 @@
 #define __RTL8372_COMMON_H__
 
 #include <linux/of_mdio.h>
+#include <linux/netdevice.h>
 #include <linux/regmap.h>
+#include <linux/spinlock.h>
 #include <linux/workqueue.h>
 #include <linux/debugfs.h>
 #include <net/dsa.h>
@@ -54,6 +56,26 @@ struct rtl837x_mib_counter {
 struct rtl837x_sdsmode_map {
 	rtk_sds_mode_t mode;
 	const char *name;
+};
+
+struct rtl837x_mib_snapshot {
+	u64 rx_octets;
+	u64 tx_octets;
+	u64 rx_ucast_pkts;
+	u64 rx_mcast_pkts;
+	u64 rx_bcast_pkts;
+	u64 tx_ucast_pkts;
+	u64 tx_mcast_pkts;
+	u64 tx_bcast_pkts;
+	u32 tx_discards;
+	u32 collisions;
+};
+
+struct rtl837x_port_stats {
+	spinlock_t lock;
+	struct rtnl_link_stats64 stats;
+	struct rtl837x_mib_snapshot snapshot;
+	bool snapshot_valid;
 };
 
 typedef struct rtl837x_pnswap_cfg_s {
@@ -129,6 +151,9 @@ struct rtk_gsw {
 
 	int default_work_delay_ms;
 	struct delayed_work status_check_work;
+	struct delayed_work stats_work;
+	struct rtl837x_port_stats port_stats[RTK_MAX_NUM_OF_PORT];
+	bool stats_work_stopping;
 };
 
 extern int rtl8372n_hw_init(struct rtk_gsw *gsw, rtl837x_pnswap_cfg_t swap_cfg);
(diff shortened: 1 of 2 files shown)


MNeroba:
Publication note: During the final source audit I corrected the MIB references to the actual RTL8373 paired `_H` identifiers. The cached path now has explicit reset detection and synchronous delayed-work teardown; it does not read hardware from `get_stats64`. A full OpenWrt package build was not possible on the current macOS host because its prerequisite check cannot be satisfied.

MNeroba:
Review follow-up: fixed reset accounting. Only backwards movement of the known 64-bit MIB counters establishes a new baseline; accumulated Linux statistics are never cleared. The known 32-bit discard/collision counters use modulo-2^32 subtraction, so a normal wrap is counted as a delta. Failed reads still leave the previous snapshot and accumulated values untouched.

perceival:
Since the PR notes a local compile wasn't possible on your end, here's what a build against this tree shows: the PR head as submitted doesn't compile. `rtl837x_stats_64bit_reset_detected()`'s second parameter is named `current`, which collides with the kernel's `#define current get_current()` macro (pulled in transitively via this file's `<linux/etherdevice.h>` -> `if_ether.h` -> `skbuff.h` -> ... -> `asm/current.h` include chain). This turns the parameter declaration and the call site into a type/syntax error:

```
rtl837x_dsa_ops.c:275:57: error: function declaration isn't a prototype [-Werror=strict-prototypes]
rtl837x_dsa_ops.c:298:53: error: passing argument 2 of 'rtl837x_stats_64bit_reset_detected' from incompatible pointer type [-Wincompatible-pointer-types]
```

Renaming the parameter to e.g. `latest` builds clean locally (confirmed, produces `rtl837x_dsa.ko`).

One more thing worth folding in while you're in there: #67 (just merged) added `rtl837x_read_ethtool_stat()`, an errno-translating wrapper around `rtk_stat_port_get()` that does exactly what this PR's `rtl837x_read_stat_value()` does — worth reusing that one instead of adding a second near-identical helper.

Holding this one out of this review pass until it builds clean; happy to take another look once updated.

MNeroba:
Thanks for the detailed build report. Fixed in `2ce8337a6856`:

- renamed the reset detector parameter from `current` to `latest` to avoid the kernel `current` macro;
- removed the duplicate `rtl837x_read_stat_value()` helper and reused #67's `rtl837x_read_ethtool_stat()` wrapper.

I reran `git diff --check` and `scripts/checkpatch.pl --no-tree --terse`; both pass. The new head is pushed and Kernel CI is now running on it.

PR #67, opened 2026-09-02
rtl837x: expose standard ethtool MAC and RMON statistics

MNeroba:
## Summary

Expose standard Ethernet PHY, MAC, control, and RMON statistics for RTL8373
through the DSA ethtool statistics callbacks.

## Details

- map the existing RTL8373 `dot3StatsSymbolErrors` MIB counter to the standard
  `SymbolErrorDuringCarrier` field;
- report standard packet, octet, collision, deferred-transmission, and
  control-opcode counters;
- map the RTL8373 receive/transmit size histograms to the standard RMON
  ranges;
- leave FCS/alignment fields unset because the chip exposes them as one
  combined counter rather than two exact standard fields;
- leave the 1519-to-maximum bucket unset because this SDK does not expose the
  maximum frame size needed to describe it;
- use the existing RTK/DAL MIB path with the valid paired `_H` identifiers for
  full-width counters.

The PHY symbol-error mapping follows the Linux structured ethtool PHY
statistics field and uses only a counter explicitly exposed by the RTL8373
SDK. No synthetic, inferred, or undocumented PHY metrics are added.

The callbacks return only values that can be represented by the standard
ethtool structures; they do not invent unsupported ranges or split a
combined error counter into duplicate fields.

## Validation

- `git diff --check`: passed.
- Confirmed `dot3StatsSymbolErrors` is declared by `mib.h` and is the exact
  counter read by the new callback.
- Confirmed Linux 6.18 DSA provides `get_eth_phy_stats` and the upstream
  Realtek RTL8365MB driver maps its symbol-error MIB to
  `SymbolErrorDuringCarrier`.
- Corrected the original MIB references to the identifiers actually declared
  by `mib.h` and made the helper name unique to avoid collision with the
  cached `get_stats64` implementation.
  current macOS host because OpenWrt rejects Apple's system `make` as
  unsupported.
- No hardware test was run.

This edits the same DSA operations file as the cached `get_stats64` patch;
the two PRs have only a textual/integration relationship and should be
rebased as needed when #66 is merged.

### DIFF
--- a/package/kernel/rtl837x/src/rtl837x_dsa_ops.c
+++ b/package/kernel/rtl837x/src/rtl837x_dsa_ops.c
@@ -179,11 +179,25 @@ static int rtl837x_set_stp_state(struct rtk_gsw *gsw, int port, u8 state)
 	return rtl837x_to_errno(ret);
 }
 
+static int rtl837x_read_stat_value(int port, rtk_stat_port_type_t counter,
+				   u64 *value)
+{
+	rtk_stat_counter_t counter_value = 0;
+	int ret;
+
+	ret = rtk_stat_port_get(port, counter, &counter_value);
+	if (ret)
+		return rtl837x_to_errno(ret);
+
+	*value = counter_value;
+	return 0;
+}
+
 static u64 rtl837x_read_stat(int port, u32 counter)
 {
-	rtk_stat_counter_t value = 0;
+	u64 value;
 
-	if (rtk_stat_port_get(port, counter, &value))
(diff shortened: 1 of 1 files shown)


MNeroba:
Publication note: During the final source audit I corrected the MIB references to the actual RTL8373 paired `_H` identifiers and gave this patch its own helper name so it can coexist with the cached statistics work. Combined FCS/alignment and unspecified 1519-to-maximum ranges remain unset rather than being reported inaccurately. A full OpenWrt package build was not possible on the current macOS host because its prerequisite check cannot be satisfied.

PR #68, opened 2026-09-02
qualcommbe: ipq53xx: drop obsolete wireless retry workaround

MNeroba:
## Summary

Remove the obsolete delayed wireless retry workaround from the IPQ53xx target.

## Why this is correct

The generic wifi-scripts change `ca476d3ddd8f74ea3572d2c1d05b0cde55d2b12e` (`wifi-scripts: retry setup after a late PHY appears`) now covers this failure class. It notices a PHY appearing after the initial netifd wireless setup, recognizes `retry_setup_failed`, and queues the retry through the generic state machine so setup/teardown races are handled there.

The target-specific script waited ten seconds and re-ran the entire `wifi up` path whenever any radio reported `retry_setup_failed`. With the generic retry in place, that hook is redundant and can repeat unrelated setup. The separate `40-gl-wireless-country` handler remains in the current base for regulatory country data for late radios, but it is context for a different concern, not the direct replacement for this retry hook.

## Validation

- Rebased directly onto current `upstream/flint3-be9300` at
- Confirmed the generic `retry_setup_failed` state machine and
  `40-gl-wireless-country` handler remain in the base.
- Final diff is deletion-only: the target-specific
  `99-retry-setup-failed` hook.
- `git diff --check`: passed.
- The change is deletion-only; no runtime hardware test was run.
- Full OpenWrt target validation was not available on the current macOS host
  because the prerequisite check rejects the Apple-provided GNU Make 3.81
  before compilation.

This PR does not change the generic retry state machine, the current country
hotplug handler, or the normal wireless setup sequence.


MNeroba:
Publication note: I rechecked the deletion against the current IPQ53xx wireless lifecycle. The targeted late-PHY country handler remains in place; this removes only the broad delayed `wifi up` retry, including the repeated retry of permanently invalid ART state. The change is deletion-only.

MNeroba:
Rationale follow-up: the description now identifies `ca476d3ddd8f74ea3572d2c1d05b0cde55d2b12e` (`wifi-scripts: retry setup after a late PHY appears`) as the generic replacement that makes this target-specific retry hook redundant. The GL.iNet country handler is mentioned only as separate context, not as the replacement for the retry state machine. No code was restored or added.

PR #69, opened 2026-09-02
rtl837x: align and enable link-down FDB age-out

MNeroba:
## Summary

Align and enable the RTL8373 chip-global link-down FDB age-out operation.

## Details

The RTL8373 DAL controls one chip-global link-down flush setting and does not
consume a port argument. The RTK wrapper previously accepted a port, rejected
only an out-of-range value, and then discarded it before calling the same
global mapper operation. Remove that misleading argument and update the
wrapper documentation while leaving the DAL implementation and mapper wiring
unchanged.

The switch setup now enables this existing DAL capability once, immediately
after `rtk_l2_init()`, through the corrected global RTK API:
`rtk_l2_flushLinkDownPortAddrEnable_set(ENABLED)`. A setup failure is
propagated. No link-state callback, per-port write, or teardown write is
added; the setting is chip-global and switch reset/reinitialization owns the
hardware state.

## Current-base integration

The branch is rebased directly onto the current `flint3-be9300` tip
`15c0c8175c846576bea17e2aa56354fdc5f6121e`. Merged #63 learning, #64
hairpin, #65 isolation, #66 statistics, #67 ethtool statistics, and #71 FDB
flush-mapper changes are inherited from the base. No duplicate or unrelated
callback changes were added while resolving the rebase.

## Validation

- `git diff --check`: passed.
- `scripts/checkpatch.pl --no-tree --terse`: passed with no output/errors for
  the final diff.
- Final diff is limited to the RTK public wrapper/header signature correction
  and one setup call; the existing DAL implementation and mapper wiring are
  unchanged.
- Verified all RTL837x call sites and DAL prototypes use the new global
  signature.
- A full `make package/kernel/rtl837x/compile V=s` is not available on this
  macOS host because OpenWrt rejects Apple's GNU Make 3.81 before package
  compilation.
- No hardware test was run.

## Dependencies

No hard code dependency remains. This PR is independent of the open bridge,
FDB, policing, and MST work.

### DIFF
--- a/package/kernel/rtl837x/src/rtk-api/l2.c
+++ b/package/kernel/rtl837x/src/rtk-api/l2.c
@@ -574,9 +574,8 @@ rtk_api_ret_t rtk_l2_table_clearStatus_get(rtk_l2_clearStatus_t *pStatus)
 /* Function Name:
  *      rtk_l2_flushLinkDownPortAddrEnable_set
  * Description:
- *      Set HW flush linkdown port mac configuration of the specified device.
+ *      Set HW flush linkdown port mac configuration.
  * Input:
- *      port - Port id.
  *      enable - link down flush status
  * Output:
  *      None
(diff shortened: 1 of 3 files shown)


MNeroba:
Publication note: I checked every RTL837x reference and the DAL mapper signature. The hardware operation is chip-global, so the old public `port` parameter was only misleadingly validated and then discarded; the DAL implementation and mapper wiring remain unchanged.

PR #70, opened 2026-09-02
rtl837x: add bridge flag regression test coverage

MNeroba:
## Summary

Add a source-tree manual regression test for RTL837x bridge-port flag handling.

## Coverage

`package/kernel/rtl837x/tests/bridge-flags.sh --manual` checks readback for learning, unicast flood, multicast flood, broadcast flood, hairpin, and isolated flags. It then prompts for physical forwarding checks covering learning, flood behavior, same-port hairpin reflection, isolated-port separation, CPU reachability, and isolated-to-normal forwarding.

The script snapshots bridge membership and per-port VLAN output, restores every original flag on exit, and refuses to report a clean run if membership or VLAN state changed. Hardware forwarding checks are explicit `pass`/`fail`/`skip` prompts; an incomplete manual run exits with status 77 rather than success.

## Hardware topology

For the hairpin check, use a true Ethernet hub/repeater as the simple
topology:

```text
host A ---\
           HUB ---- PORT_A on Flint
host B ---/
```

An ordinary unmanaged switch is not a valid generic substitute because it
may deliver host A to host B locally without sending the frame to RTL8373.
A managed downstream switch is acceptable only when its host ports are
explicitly prevented from forwarding directly to each other and traffic must
use the Flint uplink. Before toggling hairpin, have both hosts send traffic
toward Flint and confirm that the RTL8373 FDB learned both host MACs on
`PORT_A`; then verify ingress `PORT_A` to egress `PORT_A` is blocked with
hairpin off and allowed with it on.

The isolated-port checks run in both directions (`PORT_A -> PORT_B` and
`PORT_B -> PORT_A`), while retaining checks for CPU reachability and
isolated-to-normal-port forwarding. The multicast check uses a raw L2 frame
with destination `01:00:5e:7f:01:23` and non-IP EtherType `0x88b5`, with no
MDB entry or IGMP/MLD join, so it tests only `BR_MCAST_FLOOD`.

## Usage

```sh
package/kernel/rtl837x/tests/bridge-flags.sh --manual BRIDGE PORT_A PORT_B NORMAL_PORT
```

The test is intentionally not installed or run automatically by the kernel package. It requires root, the `ip`/`bridge` tools, external peers, and the corresponding Linux 6.18 driver support.

## Validation

- Rebased directly onto current `upstream/flint3-be9300` at
- Final diff remains one new test file with executable mode `100755`.
- `git diff --check`: passed.
- `sh -n package/kernel/rtl837x/tests/bridge-flags.sh`: passed.
- Executable mode verified.
- Reviewed for POSIX `/bin/sh` constructs; no non-standard shell dependency was added.
- No hardware test was run from this host.

### DIFF
--- a/package/kernel/rtl837x/tests/bridge-flags.sh
+++ b/package/kernel/rtl837x/tests/bridge-flags.sh
@@ -1,0 +1,383 @@
+#!/bin/sh
+# SPDX-License-Identifier: GPL-2.0-or-later
+
+set -eu
+
+readonly RTL837X_FLAGS="learning flood mcast_flood bcast_flood hairpin isolated"
+
+usage()
+{
+	cat <<'EOF'
+Usage:
+  bridge-flags.sh --manual BRIDGE PORT_A PORT_B NORMAL_PORT
+
+The ports must already be members of BRIDGE. PORT_A and PORT_B must be
+connected to separate external hosts. For the hairpin test, connect two
+hosts behind PORT_A through an unmanaged switch or hub. NORMAL_PORT is used
+by the isolated-port test.
+
+The script needs root, ip, bridge, awk, mktemp and a Linux 6.18 DSA driver
+with the bridge flags under test. It changes only bridge-port flags. It
+does not change bridge membership or VLAN configuration, but snapshots both
+and verifies that they are unchanged during cleanup.
+
+The flag readback checks are automatic. Forwarding checks are deliberately
+interactive because they need physical peers and packet observation:
+
+  - learning: a fresh source must not/become dynamically learned;
+  - flood: an unknown unicast must not/may reach PORT_B;
+  - bcast_flood: broadcast must not/may reach PORT_B;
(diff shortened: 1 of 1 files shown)


MNeroba:
Publication note: I rechecked the test for state safety and reviewability. It restores all recorded flags, verifies bridge membership and VLAN output, requires explicit pass/fail/skip answers for physical forwarding checks, and returns 77 for an incomplete manual run. `sh -n` and executable mode validation pass; the test is intentionally not automatic.

MNeroba:
Review follow-up: corrected the hardware procedure. The hairpin instructions now require a true hub/repeater or a managed downstream switch with host-to-host forwarding disabled, require both MACs to be learned on the RTL8373 port before the reflection check, and state the off/on same-port semantics. The isolated test now checks both directions, and multicast uses a raw non-IP L2 frame without MDB/IGMP/MLD participation.

PR #87, opened 2026-09-16
qualcommbe: ppe: IPQ5332 scheduler/buffer tables, AC table and EDMA IRQ fixes

kubrickfr:
The PPE driver configures every IPQ SoC as if it were an IPQ9574. On the
IPQ5332 that starves the 2.5G WAN port, and two further bugs in the same area
-- one silent, one a potential use after free -- turned up while chasing it.
Six commits: four driver patches, an interrupt-placement patch, and a
packet-steering hook that keeps the placement from being undone.

## The WAN ingress ceiling

`ppe_config.c` programs the IPQ9574 BM (TDM) and QM scheduler tables on all
SoCs. The IPQ9574 has eight PPE ports; the IPQ5332 has three. Per-port ingress
shares under that table:

| PPE port | GL-BE9300 role | Ingress slots (of 50) |
|---|---|---|
| 0 | CPU / EDMA | 11 |
| 1 | 10GBASE-R link to the RTL8372N switch (`lan`) | 7 |
| 2 | 2.5G USXGMII WAN (`wan`) | **2** |
| 3-7 | do not exist on the IPQ5332 | 2, 2, 11, 11, 4 |

The port carrying all WAN traffic gets the smallest share of any port that
physically exists, and 3.5x less than the LAN-side link beside it -- which is
why LAN-side testing does not show this.

On a GL-BE9300 it shows as ingress drops on PPE port 2, counted one-for-one as
XGMAC `rx_fifo_overflow`, with the BM and QM drop counters at zero and the CPUs
idle:

| | |
|---|---|
| Router-local download over the 2.5G WAN | 124 Mbit/s, ~1/5 of frames lost |
| Forwarding through the WAN | ~1.9 Gbit/s, `rx_fifo_overflow` climbing |
| Stock GL.iNet firmware, same unit | 2.2 Gbit/s, zero overflows |

The XGMAC has a 1 KB Rx FIFO in cut-through mode, so it depends on the PPE
draining it without gaps; a 2-in-100 slot share cannot do that. Under the
64-bytes-per-PPE-clock model at 200 MHz, 2 slots in 100 works out at roughly
2 Gbit/s -- so the `~1.8-1.9 Gbit/s` in this repo's README, and its open "PPE
WAN RX FIFO overruns" known issue, are this ceiling.

`0375` and `0376` install the vendor SSDK MPPE tables and budgets for the
IPQ5332 behind a new `ppe_is_ipq5332()` helper, leaving the IPQ9574 path
untouched:

- **0375** -- `ppe_port_tdm0_tbl` (depth 32) and `ppe_port_scheduler0_tbl`
  (depth 10), from `src/init/ssdk_appe.c`. Port 2 goes from 2 ingress slots in
  100 to 6 in 31.
- **0376** -- the BM/QM budgets from `qca_hppe_bm_hw_init()` and
  `qca_hppe_qm_hw_init()` in `src/init/ssdk_hppe.c`: a 240-buffer shared BM
  group over 10 BM ports instead of 1550 over 15, a 500-buffer QM group instead
  of 2000, queue ceilings of 50 instead of 1200.

Both sets of values are from the public qca-ssdk source and match the shipped
GL.iNet 4.10.0 `qca-ssdk.ko` entry for entry; stock firmware also reports the
same QM values at run time.

## Two bugs found on the way

- **0377 -- multicast queue AC table access.** The table is indexed from the
  first multicast queue (as the vendor does with `obj_id - UCAST_QUEUE_ID_MAX`),
  but the driver indexes it with the absolute queue ID (256-299), landing
  0x1000 past the table -- so the multicast thresholds were never written. It
  also bulk-transfers five words per entry where the entry holds three and the
  stride is four, spilling each write into the next entry. Affects the IPQ9574
  as much as the IPQ5332.

- **0378 -- EDMA IRQ unwind on partial failure.** If `request_irq()` fails for
  any Tx completion ring, or any Rx descriptor ring after the first, the unwind
  frees the ring names but leaves the interrupts registered, with handlers
  pointing into `edma_ctx` and names that are now freed memory. The probe then
  fails and `edma_ctx` is freed, so a later interrupt on one of those lines is a
  use after free. Now counts what was actually requested and frees exactly that.

## 0379 -- spread the EDMA ring interrupts over the CPUs

EDMA interrupts are requested with no affinity, so under the kernel's default
placement every Rx descriptor and Tx completion ring is serviced by CPU0 and the
interrupt and NAPI work of the entire datapath runs on one core. The driver
already assumes otherwise -- `edma_cfg_rx_rps_hash_map()` spreads the RSS hash
over the Rx rings and notes that they "are assigned to different cores using IRQ
affinity configuration", which nothing does -- and on the Tx side ring `i` is
already paired with the Tx queue of CPU `i % num_possible_cpus()`. Stock
firmware applies the same round-robin placement from its `qca-nss-dp` init
script.

This one depends on the Tx queue wake fixes already in the tree (e2599aafc2,
dbc044abbe); without them a ring serviced on the fallback CPU would wake the
wrong queue.

## The packet-steering hook

The last commit adds `usr/libexec/platform/packet-steering.sh` for the
subtarget. Without it, `packet-steering.uc` undoes 0379: `lan` and `wan` share
one platform device and have no threaded NAPI, so it takes its single-device
path and assigns the pair one CPU. Its own dry run on a GL-BE9300:

```
echo 1 > lan/queues/rx-0/rps_cpus     echo 1 > wan/queues/rx-0/rps_cpus
echo 1 > lan/queues/rx-1/rps_cpus     echo 1 > wan/queues/rx-1/rps_cpus
echo 1 > lan/queues/rx-2/rps_cpus     echo 1 > wan/queues/rx-2/rps_cpus
echo 1 > lan/queues/rx-3/rps_cpus     echo 1 > wan/queues/rx-3/rps_cpus
```

All eight queues to mask 0x1 -- CPU0, the core already servicing `rxdesc_12`,
`txcmpl_4` and `txcmpl_8`. The hook runs the generic logic first, so the Wi-Fi
netdevs are still configured, then clears `rps_cpus` on the PPE ports. It reads
`steering_flows` itself because the init script only passes it to the generic
path.

## Testing

On a GL-BE9300, kernel 6.18, all six applied. Ring interrupts land one per core
and stay there under load:

```
rxdesc_12  443944       0       0       0   CPU0
rxdesc_13       0  836272       0       0   CPU1
rxdesc_14       0  845454       0       0   CPU2
rxdesc_15       0       0       0 1102824   CPU3
txcmpl_4/8 → CPU0   txcmpl_5/9 → CPU1   txcmpl_6/10 → CPU2   txcmpl_7/11 → CPU3
```

With the hook in place, all eight PPE rx queues read `rps_cpus = 0` while the
Wi-Fi netdevs keep theirs, across a `/etc/init.d/packet_steering reload`.

**Not yet measured:** WAN throughput and `rx_fifo_overflow` counters with
`0375`+`0376` applied. The before numbers are in the table above; the after
numbers are the point of the series and I have not captured them yet. Happy to
hold the series until I have, if you would rather see them first.

## Notes

- `0375`, `0376` and `0379` are gated on the IPQ5332 or are pure placement
  changes. `0377` and `0378` change behaviour on the IPQ9574 too, and both look
  like clear bugs on either SoC.
- Separately, `packet-steering.uc` requires `rx_tasks > 1` before it will assign
  per-queue CPUs, so any multiqueue NIC with per-queue IRQ affinity but no
  threaded NAPI gets one mask smeared across all its queues. That looks worth
  fixing upstream in OpenWrt, but it changes behaviour for every target, so it
  is not in this series.


perceival:
Thanks for this — really thorough writeup, and the ingress-slot table makes the WAN ceiling obvious in hindsight.

Build-verified locally: all 6 commits apply cleanly (fast-forward, zero fuzz) onto `flint3-be9300` @ `b442595547`, and a full kernel rebuild (`glinet_gl-be9300` target, kernel 6.18.39) compiles clean — 0 errors, no new warnings traced to `edma.c`/`ppe_config.c` beyond the pre-existing benign `missing MODULE_DESCRIPTION()` one on `qcom-ppe.o`. Both prerequisite SHAs for 0379 (`e2599aafc2`, `dbc044abbe`) are confirmed present — the latter is our own patch 0374.

0377 and 0378 read as correct, narrow fixes on inspection (the AC-table indexing/stride bug, and the EDMA IRQ-unwind UAF) — independently valuable regardless of what happens with the rest of the series.

One thing worth knowing: we already carry a *runtime* workaround for the same CPU0-EDMA-IRQ problem 0379 fixes properly at the driver level — an `/etc/rc.local` script pinning `smp_affinity` per ring, deployed live and posted as forum #161. Your fix is the right way to do this (and additionally catches something ours didn't — `packet-steering.uc` clobbering `rps_cpus` back to one CPU on non-threaded-NAPI devices). If 0379 lands we'd retire the rc.local hack.

On 0375/0376: agreed these are the ones that matter most and the ones we can't sign off on without the after-numbers you mentioned. We can't get bench numbers near-term, so happy to hold as you offered — 0377/0378 look independently landable regardless of the rest.

kubrickfr:
Thanks for the build verification — and good timing: I have the after-numbers now.

## Method

Identical to the 14 Sep baseline, same command, same URL, same counters:

```
timeout 12 wget -q -O /dev/null http://ping.online.net/1000Mo.dat
```

terminating on the router itself (no forwarding, no client involved). Rate from
the `pppoe-wan` byte counter in `/proc/net/dev`; loss from `rx_fifo_overflow` in
`ethtool -S wan` before and after.

## Router-local, single stream — the direct before/after

| | Before (14 Sep, unpatched) | After |
|---|---|---|
| Throughput | 124 Mbit/s (186 MB / 12 s) | **627 Mbit/s** (941 MB / 12 s) |
| `rx_fifo_overflow` | 26,866 | **0** |
| Frames received | ~163,000 | 648,543 |
| Loss | ~1 in 6 (16.5 %) | **0.00 %** |

5x the throughput and the overflow counter flat at zero, on a run that
previously lost a sixth of its frames.

## Pushing past the old ceiling

A single flow hashes to one Rx ring and therefore one core, so the 627 Mbit/s
above is CPU-bound, not PPE-bound — one core sat at 80 % while the other three
idled. With parallel streams:

| Test | Throughput (wan port) | Frames | `rx_fifo_overflow` | Per-CPU busy |
|---|---|---|---|---|
| Router-local, 4 streams, 16 s | 1909 Mbit/s | 2,502,359 | **0** | 51 / 87 / 92 / 43 % |
| Router-local, 8 streams, 16 s | 1968 Mbit/s | 2,580,224 | **0** | 77 / 83 / 60 / 84 % |
| Forwarding, 2.5G client, 16 s | **2015 Mbit/s** | 2,642,497 | **0** | 71 / 84 / 2 / 1 % |

The forwarding figure is the one to compare against the old `~1.9 Gbit/s with
the counter climbing`: same rate, zero drops across 2.64 M frames. `rx_pause`,
`tx_pause`, `rx_drop_frames`, `rx_crc_err` and `rx_len_err` were all zero on
every run.

The ceiling has moved off the PPE. Router-local now tops out on CPU (all four
cores 60-84 % at 1968 Mbit/s, terminating PPPoE and TCP on 4x A53), and
forwarding at roughly what this line delivers — stock firmware measured
2.2 Gbit/s PPPoE on the same unit.

## Caveats, stated plainly

- **The before-numbers are the 14 Sep run on an unpatched build, not a
  re-measurement.** Same method, same URL, same counters, same 12 s — but it is
  a cross-build comparison, not a same-session A/B. I did not re-flash an
  unpatched image to re-run it.
- **The firmware under test was built 15 Sep, from these patches while they were
  still uncommitted**, so its version string reads `r35676-b442595547` (the base
  commit) and it is not provably byte-identical to what is in this PR. I
  verified the relevant state from the hardware rather than trusting the string:

  | Register | Value | Meaning |
  |---|---|---|
  | `PPE_BM_SCH_CTRL` @ `0xb000` | `80000020` | `SCH_DEPTH` = 32 → IPQ5332 table (IPQ9574 = 100) |
  | `PPE_BM_SHARED_GROUP_CFG` @ `0x600290` | `000000f0` | `SHARED_LIMIT` = 240 → IPQ5332 budget (IPQ9574 = 1550) |
  | `/proc/irq/{48..51}/affinity_hint` | 1, 2, 4, 8 | set via `irq_update_affinity_hint()` — driver-side |

  So 0375, 0376 and 0379 are demonstrably live in the image that produced these
  numbers. If you want the numbers re-taken on a build made from this branch as
  committed, say so and I will redo them.
- The `CPUs idle` phrasing in the original 124 Mbit/s report rested on the clock
  staying at its 1.1 GHz floor, not on per-core sampling — that sampler was
  broken on that run. The per-core figures in the tables above are real
  `/proc/stat` deltas.

## On the rc.local workaround

Useful to know, thanks. Incidentally that distinction is what let me confirm
0379 was actually live rather than a leftover script: a shell script writing
`smp_affinity` cannot populate `affinity_hint`, and the hints were set — only
`irq_update_affinity_hint()` in the driver does that. So if you retire the
rc.local hack, the hint values are a reliable way to confirm the driver took
over.

## On splitting

Agreed on 0377/0378 — they are independent of the IPQ5332 work and stand on
their own, so please take them separately if that is easier. Happy to resubmit
them as their own series if you would rather keep this PR scoped to the
scheduler/buffer tables.

perceival:
Merged — thank you both for the writeup and the follow-through on the numbers.

Before merging we cross-checked the committed patches against the real vendor `qca-ssdk` source (`ssdk_appe.c`/`ssdk_hppe.c`, `MPPE_TYPE` case) entry-for-entry, not just the register-level scalars you measured: `ppe_port_tdm0_tbl`/`ppe_port_scheduler0_tbl` table *contents* (not just depth), and every BM/QM budget field, match the vendor source exactly. That closes the "built before commit" provenance gap you flagged — the SCH_DEPTH/SHARED_LIMIT/affinity_hint readback you did nailed the scalars, and the table-content diff confirms the rest.

One thing worth a heads-up: `0377`'s AC-table indexing fix touches shared code, so it also affects the `ipq95xx` (IPQ9574) target, not just `ipq53xx`. Might be worth flagging separately or upstreaming on its own, as you mentioned.

We'll retire the `/etc/rc.local` `smp_affinity` workaround now that 0379 is in — `affinity_hint` being populated (driver-only, as you noted) is the confirmation the hack is superseded.

PR #92, opened 2026-09-21
qualcommbe: ppe: fix the Rx buffer DMA mapping lifecycle

kubrickfr:
Patches 0420–0421.

### 0420 — direction mismatch

`edma_rx_alloc_buffer_list()` maps the buffers it posts with `DMA_TO_DEVICE`, in both the linear and the page-mode branch, while `edma_rx_reap()` unmaps them with `DMA_FROM_DEVICE`. The device writes these buffers and the CPU reads them, and nothing writes the mapped region from the CPU between the mapping and the producer-index update, so `DMA_FROM_DEVICE` is right for both halves. The lengths already agree.

**Stated honestly in the commit message:** nothing is known to misbehave on this SoC, where the arm64 implementation cleans the buffer on map whichever direction it is given and invalidates on unmap. This is an API-correctness fix — `dma_debug` reports the mismatch, and behind an IOMMU `dma-iommu` derives the mapping permissions from the direction, so `DMA_TO_DEVICE` maps the buffer without device write permission and the device's write faults.

### 0421 — teardown leaks a mapping per in-flight buffer

Both Rx cleanup paths walk the descriptors that still hold a buffer and free the skb behind each one **without unmapping first**; `edma_rx_reap()` is the only unmap site. Teardown therefore leaks a streaming mapping for every buffer in flight, and in page mode hands the page back through the skb fragment while it is still mapped. What leaks close to a full ring per fill ring is an unwind from a failed `edma_hw_configure()` after the rings are filled, and every module unload.

The fill descriptor had no accessor for its buffer address, so one is added next to the setter; `word0` is byte-swapped by `EDMA_RXFILL_ENDIAN_SET()` and the new accessor swaps it back the way `EDMA_RXDESC_BUFFER_ADDR_GET()` does.

### Verification

Static plus review — these are the least-exercised patches in the set. 0421's unmaps *do* now execute: module unload completes on hardware (see the `ppe/lifetime-fixes` PR and the rtl837x PR), where previously it hung before ever reaching ring cleanup, so this code was unreachable in practice.

### Notes and caveats

- 0421 follows 0420 so the unmap directions it adds match the corrected map direction.
- 0421 has a **value** dependency on the `ppe/lifetime-fixes` PR: before the NAPI teardown fixes, unload never got far enough for these unmaps to run. No textual dependency.
- Documented in 0421 and not fixed here: the rings are only disabled, not drained, before cleanup runs, so a buffer the device is still writing is unmapped underneath it, and one taken from the fill ring but not yet produced into the descriptor ring is in neither walk. That wants a separate look at the ring-disable path — tracked in the follow-up issue.
- `edma_cfg_rx_rings_cleanup()` cleans the fill rings before the descriptor rings, which now read their fill ring's geometry. Safe because fill cleanup only releases descriptor memory and the rings are freed after both loops — stated in the commit message since the ordering does not show it.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

---

### Part of a five-PR series

| branch | scope |
|---|---|
| `rtl837x/conduit-refcount` | DSA conduit netdev reference (independent of the rest) |
| `ppe/lifetime-fixes` | probe/remove hangs and use-after-frees |
| `ppe/rx-queue-map` | PPE RING_Q_MAP per-ring queue sets |
| `ppe/runtime-knobs` | RPS sysctl and mitigation timer module parameters |
| `ppe/rx-dma-mapping` | Rx buffer DMA direction and teardown unmap |

Suggested order: `rtl837x/conduit-refcount` (independent) → `ppe/lifetime-fixes` → `ppe/rx-queue-map` and `ppe/rx-dma-mapping` → `ppe/runtime-knobs`.

Patch numbers are disjoint across the series so merge order cannot collide on filenames; whichever lands second may want a `make target/linux/refresh` to absorb line-number drift. Each branch was regenerated against a pristine tree — every patch applies with zero fuzz and the driver compiles clean on each branch independently.


kubrickfr:
Hi @perceival , would you mind dropping me an email at kubrick_at_fgv6_dot_net please so we can have a chat? Or propose another communication channel. Thanks!

